PR #41, opened 2025-02-21
Fix(spec): elaborate on treasury Withdraw redeemer

euonymos:


specification/chapters/2-L1-rule-based-regime.tex
```diff
@@ -365,6 +373,13 @@ \section{Treasury}%
             \mathcal{RH}_{32} \; \T{UtxoSet^{L2}}
         \end{split}
         \end{equation*}
+      \item All \code{withdrawnUtxos} are outputs of the transaction:
+         \begin{equation*}
+         \begin{split}
+            &\T{withdrawnUtxos} :: [Output^{L2}] \\
+            &\T{withdrawnUtxos} \coloneq \; \T{map} \; \T{utxo} \; \T{withdrawals}
+         \end{split}
+         \end{equation*}
```
GeorgeFlerovsky: I'll combine this with the previous step's iteration over `withdrawals`.

Also, it seems to be easier to describe the iteration as a for-each loop with mutable variables than a fold.

GeorgeFlerovsky (approved):
Thanks for your suggestions in this PR. I reworked them a bit and I think the spec has improved as a result.

PR #78, opened 2025-03-11
Wiring minimal L2 ledger, block production 

euonymos:
- [x] Refactor state mgmt
- [x] Wire-in L2 ledger
- [x] Awaiting deposits -> L1 state
- [x] Generalize block production
- [x] Type-safe block builder
- [x] Revise and re-wire block effects
- [x] API for L2 txs (and withdrawals)
- [x] Add withdrawals to settlement txs
- [x] Add withdrawals to finalization txs
- [x] Visualize both L1 and L2
- [x] Virtual genesis txs
- [x] Virtual withdrawal txs
- [x] Use virtual genesis tx hash as txId for genesis
- [x] Use `type OutputRef = (TxId, TxIx)`
- [x] Rename Refund.scala -> RefundTx.scala and other tx builder files
- [x] ~~Use `depositWindow` in lieu of `deadline` (see https://github.com/cardano-hydrozoa/hydrozoa/pull/52#discussion_r1986584376)~~ moved to #88
- [x] ~~Use static Network id/magic (all heads)~~ moved to #84 
- [x] Separate the pure part of block creation
- [x] remove --accept-flake-config
- [x] Separate tx logs for L1 and L2


src/main/scala/hydrozoa/node/server/Node.scala
```diff
@@ -207,78 +216,187 @@ class Node(
...
-
-        // TODO: temporary: handle the event
-        multisigEventManager.map(_.handleFinalizationTx(finalizationTx, finalizationTxId))
-
-        Right(serializedTx)
+    /** Manually triggers next block creation procedure.
+      * @param nextBlockFinal
+      * @return
+      */
+    def handleNextBlock(nextBlockFinal: Boolean): Either[String, String] =
```
GeorgeFlerovsky, line 231: I would prefer the logic of handleNextBlock to live as a pure function in `Block.scala`. I would pass `finalizing`, `previousBlock`, `previousMajorBlock`, `stateL2.utxosActive`, and `timeCurrent` into it as (immutable) function arguments.

Is there a reason why you chose to put it into `Node.scala`?

src/main/scala/hydrozoa/node/server/State.scala
```diff
@@ -1,166 +1,252 @@
...
+            case None => foo(HeadAbsentStateImpl())
+            // Should be this: (initializing state is missing now).
+            // case Some(Initializing(_)) => foo(HeadAbsentStateImpl())
+            case _ =>
+                throw IllegalStateException("The head is missing or not in Initializing state.")
+
+    def asOpen[A](foo: OpenNodeState => A): A =
+        headState match
+            case Some(open: Open) => foo(OpenNodeStateImpl(open))
+            case _ => throw IllegalStateException("The head is missing or in a Open state.")
```
GeorgeFlerovsky, line 200:
```suggestion
            case _ => throw IllegalStateException("The head is missing or not in a Open state.")
```

src/main/scala/hydrozoa/node/server/Node.scala
```diff
@@ -207,78 +216,187 @@ class Node(
...
+        val prevHeader = state.asOpen(_.l2Tip.blockHeader)
+
+        // (c) Let previousMajorBlock be the latest major block in blocksConfirmedL2
+        //val previousMajorBlock = state.asOpen(_.l2LastMajor)
+
+        // (d) Let utxosActive be a mutable variable initialized to stateL2.utxosActive
+        // for now (and probably this is legit) we use utxosActive within L2 ledger
+        // var utxosActive: UTxOs = state.asOpen(_.utxosActive)
+        // TODO: do we need to clone the ledger for block creation?
+        val stateL2 = state.asOpen(_.stateL2)
```
GeorgeFlerovsky, line 251: The head state's confirmed L2 active utxo set should _never_ be modified when you create a new block.
The L2 effect of a new block only happens when it is confirmed -- all peer signatures have been collected.

Unless you have another way to achieve this, I think you do need to clone the L2 active utxo set.

GeorgeFlerovsky (changes requested):
Quick high-level review

src/main/scala/hydrozoa/node/server/Node.scala
GeorgeFlerovsky: Theoretically, we could get fancy with some "copy on write" virtualization that reuses the confirmed L2 active utxo set (immutably) and overlays the added/deleted L2 utxos on top of it...

GeorgeFlerovsky: On the other hand, we are already keeping track of the `utxosAdded` and `utxosWithdrawn`, so perhaps the overlay is not so far fetched...

src/main/scala/hydrozoa/node/server/Node.scala
euonymos: Yep, just put it where I was. Yes, I like the idea of splitting up a "pure" part. I use quotes since Scala is not pure as a language, so we need to maintain purity by not passing things like `NodeStateManager`.

src/main/scala/hydrozoa/node/server/Node.scala
GeorgeFlerovsky: For extra efficiency, `utxosRemovedTxs` should only track the utxos spent from `utxosActiveConfirmed`. When utxos from `utxosAddedTxs` are spent, we can just directly remove them from that set.

GeorgeFlerovsky: For efficiency, when L2 transactions (but not withdrawals!) spend utxos from `utxosAddedTxs`, we can just directly remove those utxos from that set.

src/main/scala/hydrozoa/node/server/Node.scala
euonymos: Done.

GeorgeFlerovsky (approved):
Mostly LGTM. See comments

src/main/scala/hydrozoa/node/server/State.scala
```diff
@@ -1,166 +1,251 @@
...
+    def immutablePoolEventsL2: Seq[L2NonGenesis]
+    def immutableBlocksConfirmedL2: Seq[Block]
+    def immutableEventsConfirmedL2: Seq[(L2Event, Int)]
+    def enqueueDeposit(deposit: DepositUtxo): Unit
+    def poolEventL2(event: L2NonGenesis): Unit
+    def stateL1: MultisigHeadStateL1
+    def stateL2: AdaSimpleLedger[THydrozoaHead]
+    def l2Tip: Block
+    def l2LastMajor: Block
+    def finalizing: Boolean
```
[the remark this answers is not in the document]
GeorgeFlerovsky: We discussed this: FinalizingState is probably not needed.

Instead, we might need FinalizedState, which handles any leftover unabsorbed deposits after the head is finalized.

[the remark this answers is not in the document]
GeorgeFlerovsky: We discussed this today. The plan is to more cleanly separate node state from head state, along with some more drastic refactors.

src/main/scala/hydrozoa/infra/Tx.scala
```diff
@@ -90,22 +98,122 @@ def addWitness(tx: L1Tx, wit: TxKeyWitness): L1Tx = {
...
-def txInputsRef(tx: L1Tx): Set[(TxId, TxIx)] =
+// TODO: unused
+def txInputs[L <: AnyLevel](tx: TxAny): Set[OutputRef[L]] =
     val tx_ = Transaction.deserialize(tx.bytes)
-    tx_.getBody.getInputs.asScala.map(ti => (TxId(ti.getTransactionId), TxIx(ti.getIndex))).toSet
+    tx_.getBody.getInputs.asScala
+        .map(ti => OutputRef[L](TxId(ti.getTransactionId), TxIx(ti.getIndex)))
+        .toSet
+
+def toBloxBeanTransactionOutput(output: TxOut): TransactionOutput =
```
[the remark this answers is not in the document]
euonymos: I stopped using Scalus' types in args, so now it's `toBloxBeanTransactionOutput[L <: AnyLevel](output: Output[L]): TransactionOutput` and doesn't require access to the `Network` anymore. But this is a good observation, we need to exercise some discipline. In principle, we can use an annotation to mark pure functions.

src/main/scala/hydrozoa/l2/block/Block.scala
```diff
@@ -22,18 +30,141 @@ enum BlockTypeL2:
...
+/** We don't add genesis events to blocks, since they can't be invalid and because they can be
+  * calculated from `depositsAbsorbed`.
+  */
+enum MempoolEventTypeL2:
+    case MempoolTransaction
+    case MempoolWithdrawal
+
+type UtxoSetL2 = Utxos
+
+opaque type RH32UtxoSetL2 = H32[UtxoSetL2]
```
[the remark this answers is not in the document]
euonymos: Do you think it's so different that we need a separate type? Isn't attaching a phantom type enough? Anyway, let me move into a separate package. The opaque type is an opaque type after all we can alter it afterward.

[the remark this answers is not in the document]
euonymos: Added in dedd5b09b35d7e81e585a829bee6553a89d96dbd.

src/main/scala/hydrozoa/l2/ledger/AdaSimpleLedger.scala
```diff
@@ -1,119 +1,222 @@
...
-
-    private def eventHash(s: String): TxId =
-        TxId(encodeHex(CryptoHash.H32.hash(IArray.from(s.getBytes)).bytes))
-
-    def event(hash: L2EventHash): Option[L2Event] = ???
+object AdaSimpleLedger:
+    def apply(): AdaSimpleLedger[THydrozoaHead] = AdaSimpleLedger[THydrozoaHead](NoopVerifier)
 
-    def allEvents: Set[L2EventHash] = ???
+    def adopt(event: SimpleGenesis | SimpleTransaction | SimpleWithdrawal): (TxL2, L2EventHash) =
```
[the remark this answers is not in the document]
euonymos: Done.

src/main/scala/hydrozoa/l2/ledger/Ledger.scala
```diff
@@ -9,12 +12,20 @@ package hydrozoa.l2.ledger
   * @tparam V
   *   type for event verifier
   */
-trait L2Ledger[+U, +E, H, -V]:
+trait L2Ledger[U, G, T, W, UD, E <: AnyL2Event[H, G, T, W, UD], H, -V]:
```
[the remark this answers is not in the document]
euonymos: Done in 306db733711490d1e72fda369229a27097c99bbb.

PR #87, opened 2025-03-16
Virtual withdrawals txs for visualization

euonymos:
Saving this for later use.

The problem with this approach stems from the fact that it is very easy to add "virtual input" to a tx and impossible to add a "virtual output". Since an extra input just represents a link from a spent output to the tx, but once you add an additional output, it gets rendered as a separate node and a link to it due to a different tx hash. So, for combined L1/L2 diagrams, I decided to go back to "virtual settlement txs," which virtually consume L2 withdrawn inputs:
image

For reference: alternative approach with virtual withdrawals breaks because of the way sc-tools work now:
image


euonymos:
See #93.

PR #91, opened 2025-03-17
Address comments in #60

euonymos:


src/main/scala/hydrozoa/infra/Tx.scala
```diff
@@ -104,14 +104,13 @@ def addWitness[L <: AnyLevel](tx: Tx[L], wit: TxKeyWitness): Tx[L] = {
...
-    val tx_ = Transaction.deserialize(tx.bytes)
-    val outputs = tx_.getBody.getOutputs.asScala
-    outputs.indexWhere(output => output.getAddress == address.bech32) match
-        case -1 => None
-        case i  => Some((TxIx(i), BigInt.apply(outputs.apply(i).getValue.getCoin.longValue())))
+def onlyOutputToAddress(tx: TxAny, address: AddressBechL1): Option[(TxIx, BigInt)] =
+    val outputs = Transaction.deserialize(tx.bytes).getBody.getOutputs.asScala.toList
+    outputs.filter(output => output.getAddress == address.bech32) match
+        case List(elem) => Some((TxIx(outputs.indexOf(elem)), elem.getValue.getCoin.longValue()))
+        case _          => None
```
GeorgeFlerovsky, line 113: No, I think this should give an error if there's more than one element, because then you're likely making a mistake in downstream code that is expecting only a single element to fit the filter criteria.

GeorgeFlerovsky: The function has three cases:
```
  [a] -> Right a
  [] -> Left NoMatch
  [a,b,...] -> Left TooManyMatches
```

src/main/scala/hydrozoa/l1/multisig/tx/settlement/BloxBeanSettlementTxBuilder.scala
```diff
@@ -30,10 +29,10 @@ class BloxBeanSettlementTxBuilder(
         r: SettlementRecipe
     ): Either[String, SettlementTx] =
 
-        val refs = r.deposits.toSet + headStateReader.currentTreasuryRef
+        val inputsRefs = r.deposits.toSet + headStateReader.currentTreasuryRef
```
GeorgeFlerovsky, line 32: Still confusing terminology lol. It's quite unfortunate that Cardano's core devs, in their infinite wisdom, chose to use the word "reference" both for the unique identifier of a utxo and for a utxo that a transaction accesses in a read-only manner.

How about this:
- `UtxoId` = `OutputReference`
- `inputIds :: Set UtxoId` = output references of the transaction's inputs
- `refInputIds :: Set UtxoId` = output reference of the transaction's reference inputs

Correspondingly:
- `Utxo` = `Output`
- `UtxoSet` = `Map UtxoId Utxo`

(Older Cardano folks might get confused because they used to refer to the active utxo set as UTxO, but I don't care. Nowadays, in conversation, I think everyone says "utxo" to mean a single transaction output and "utxo set" to mean a set of utxos tagged by their ids.)

src/main/scala/hydrozoa/l1/multisig/tx/deposit/BloxBeanDepositTxBuilder.scala
GeorgeFlerovsky, line 1: - `DepositTxBuilderBloxbean.scala`
- `FinalizationTxBuilderBloxbean.scala`
- `InitializationTxBuilderBloxbean.scala`
- `RefundTxBuilderBloxbean.scala`
- `SettlementTxBuilderBloxbean.scala`

GeorgeFlerovsky (changes requested):
Sorry for the pedantic comments, but I guess that's the theme of this PR anyway. :sweat_smile:

src/main/scala/hydrozoa/infra/Tx.scala
euonymos: Yeah, I was about to do exactly that, let me do that ;-)

src/main/scala/hydrozoa/l1/multisig/tx/settlement/BloxBeanSettlementTxBuilder.scala
euonymos: Ok, I love your idea. These references always confuse me as well, I will be happy not to use them at all.

src/main/scala/hydrozoa/infra/Tx.scala
euonymos: Done.

GeorgeFlerovsky (approved):

PR #97, opened 2025-03-21
L1 mock

euonymos:
As a step towards model-based testing, we are adding a mocked L1 implementation which does minimal checks.

Happy-path scenario on:
* mocked L1: 2.5 sec
* Yaci DevKit: 7.6 sec


protocolParameters.json
GeorgeFlerovsky, line 1: Let's not pollute the root project folder with too many files.

Plz move to `infra/config/protocolParameters.json`

GeorgeFlerovsky (approved):
LGTM. Good start to mock + testing

Issue #25, opened 2025-02-18
Treasury spending validator


euonymos:
Redeemers:
* Resolve
* Withdraw
* Deinit

PR #125, opened 2025-05-30
On-chain scripts for rule-based regime

euonymos:
- [x] Datums and redeemers
- [x] BLS12-381 scalar calculations
- [x] BLS accumulators
- [x] Treasury spending script - #25 
- [x] Dispute resolution script - #24 
- [x] Tallying stake validator - #128

### DIFF
--- a/src/main/scala/hydrozoa/l2/ledger/simple/SimpleL2Ledger.scala
+++ b/src/main/scala/hydrozoa/l2/ledger/simple/SimpleL2Ledger.scala
@@ -4,13 +4,13 @@ import cats.implicits.toBifunctorOps
 import hydrozoa.*
 import hydrozoa.infra.{Piper, decodeBech32AddressL2, plutusAddressAsL2}
 import hydrozoa.l2.ledger.*
+import scalus.builtin.ByteString
+import scalus.prelude.Option.{None as SNone, Some as SSome}
+import scalus.prelude.{Option, AssocMap, given}
+import scalus.ledger.api.v1 as scalus
 
 import scala.collection.mutable
-import scalus.builtin.ByteString
-import scalus.prelude.AssocMap
-import scalus.prelude.Maybe.{Just, Nothing}
-import scalus.prelude.Prelude.given_Eq_ByteString
-import scalus.ledger.api.v1 as scalus
+
 
 /** This object defines types and constructors for Hydrozoa's L2 ledger and contains a class that
   * implements L2LedgerModule.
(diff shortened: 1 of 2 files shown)


src/main/scala/hydrozoa/l1/rulebased/onchain/scalar/Scalar.scala
```diff
@@ -0,0 +1,117 @@
...
+    def apply(n: BigInt): Option[Scalar] =
+        if n >= 0 && n < fieldPrime then Some(new Scalar(n))
+        else None
+
+    // TODO: Why doesn't it work?
+    @Ignore
+    def apply(s: String): Option[Scalar] =
+        val n = BigInt(BigInteger(s))
+        Scalar.apply(n)
+
```
nau, line 37: This should work:
```scala
inline def apply(inline s: String): Option[Scalar] =
        val n = BigInt(s)
        Scalar.apply(n)

File: src/main/scala/hydrozoa/l1/rulebased/onchain/scalar/Scalar.scala [hunk 2]
```diff
@@ -0,0 +1,117 @@
...
+        Scalar.apply(byteStringToInteger(false, bytes))
+
+    extension (self: Scalar)
+
+        // Combining
+
+        // Adds two `Scalar` elements, ensuring the result stays within the finite field range.
+        def add(addend: Scalar): Scalar =
+            new Scalar((self.unScalar + addend.unScalar) % fieldPrime)
+
```
nau, line 52: You can use `def +(add: Scala): Scala = Scalar((self.unScalar + addend.unScalar) % fieldPrime)

src/main/scala/hydrozoa/l1/rulebased/onchain/scalar/Scalar.scala
```diff
@@ -0,0 +1,117 @@
...
+                if k == BigInt(0) then self
+                else go(self.mul(self), k - 1)
+
+            if k < 0 then zero
+            else go(self, k)
+
+        def neg: Scalar =
+            if self.unScalar == BigInt(0) then self
+            else new Scalar(fieldPrime - self.unScalar)
+
```
nau, line 99: `def unary_- :Scalar = ...

src/main/scala/hydrozoa/l1/rulebased/onchain/scalar/Scalar.scala
euonymos: Now it seems to work, IDK why.

nau: That's because Scalus compiler only support creation of `BigInt("1234")` only with `String` literals, not variables.

By changing `def apply` to `inline def apply(inline s: String)` make it a macro that inlines the `s` literal into `BigInt(s)`. This happens before Scalus compiler plugin pass, so Scalus sees `BigInt("123")` instead of `BigInt(s)` – hence it works. 

Hope this helps.

PR #136, opened 2025-07-01
Remove the beaconTokenName from the InitTxRecipe

dragospe:
Small PR to remove the beaconTokenName from the InitTxRecipe and calculate it from the seed UTxO instead.

With some more work, we may be able to do the same for the head script and address -- but I think we'd need to carry around a network ID instead, which probably isn't worth refactoring until we switch tx builders.


src/main/scala/hydrozoa/l1/rulebased/onchain/DisputeResolutionValidator.scala
dragospe, line 1: and fixing a small typo I came across

euonymos:
Yep, it was done like that before. Then I tried to figure out what should be done in builders and what should be done somewhere else. My current understanding is that no specific Hydrozoa-logic should sprout into builders - they are machines to hide the complexity of Cardano transaction building. Now this is not the case, unfortunately - we DO many things like calculating the treasury amount in builders, but I would prefer to have it somewhere else.

Your change goes well with that idea, since we use that mkToken function from Hydrozoa which is fine.

PR #140, opened 2025-07-03
Test reference script deployment

euonymos:
This adds the ability to deploy Hydrozoa reference scripts and bring the references to `mkTxBuilders` function where they are likely needed.

### DIFF
--- a/integration/src/test/scala/hydrozoa/DisputeSuite.scala
+++ b/integration/src/test/scala/hydrozoa/DisputeSuite.scala
@@ -1,7 +1,12 @@
 package hydrozoa
 
+import com.bloxbean.cardano.client.backend.api.BackendService
+import com.bloxbean.cardano.client.backend.blockfrost.service.BFBackendService
+import com.bloxbean.cardano.client.spec.Script
 import com.typesafe.scalalogging.Logger
-import hydrozoa.infra.txHash
+import hydrozoa.deploy.mkDeployTx
+import hydrozoa.infra.{serializeTxHex, toEither, txHash}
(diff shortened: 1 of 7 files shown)


dragospe:
This seems to make sense.

With these sorts of integration tests, I usually like to add "external" post-condition checks. 

In this case, I'd think it'd be good to have another peer observe independently that the ref script appears on chain. I think this test is just looking at the transaction from the perspective of the submitter -- can we also check Isabel and Julia cross check the deployments?

integration/src/test/scala/hydrozoa/DisputeSuite.scala
```diff
@@ -57,7 +62,23 @@ class DisputeSuite extends FunSuite {
...
+            val tx = mkDeployTx(backendService, peer, script)
+            log.info(s"deployment tx is: ${serializeTxHex(tx)}")
+            backendService.getTransactionService.submitTransaction(tx.bytes).toEither match
+                case Right(txId) =>
+                    log.info(s"$txId")
+                    UtxoIdL1.apply(TxId(txId), TxIx(0))
+                case Left(err) =>
+                    log.error(s"Can't deploy reference scripts: $err")
+                    throw RuntimeException()
+        }
```
dragospe, line 79: This looks feasibly correct, but two comments:

- It'd be nice to have more specific exceptions during testing
- We might be able to fit both reference scripts in a single transaction. This might cut down L1 fees and deployment latency.

I don't think this is high priority, personally. But we can make a ticket and/or add a comment here.

euonymos: Added the `err` to the exception and opened #141. 

PS. Currently the scripts don't fit into one regular tx, so I doubt they will fit into one tx, but we can try.

PR #146, opened 2025-07-09
Remove ledger module 

euonymos:
In today's session, we decided to take a look at the L2 ledger component:
1. We agreed that the attempt to abstract the ledger is premature and should be avoided in favor of simplicity.
2. Then we noted that the ledger is used only when we are producing/verifying a block; in other cases, we just keep a cold utxo set.
3. So now `stateL2` is the state, not the ledger as it was before. We create a ledger every time we need it.
4. Most `type` definitions now can be safely removed along with the module, leaving a bare `SimpleL2Ledger` that can be replaced by a full-fledged L2 ledger as the next step.

cc @GeorgeFlerovsky


src/test/scala/hydrozoa/l2/ledger/SimpleLedgerSpec.scala
dragospe, line 1: Replaced with HydrozoaMutatorSpec

PR #152, opened 2025-07-15
Peter/txbuilder

dragospe:
Overall note: I'm looking for a way to reify our patterns in scala, especially to make it easy to split up work according to compiler-enforced interfaces (Alice writes queries, Bob writes calculations, Charlie writes an Initializer, Danielle writes the fixedPoint (Finalizer?)). But if this comes down to being more ergonomic just by relying on developer discipline, I'm not too attached


src/main/scala/hydrozoa/l1/rulebased/tx/resolution/ScalusResolutionTxBuilder.scala
dragospe, line 1: Ignore this file, it was from a careless `git add`

src/main/scala/hydrozoa/infra/TxBuilder.scala
```diff
@@ -0,0 +1,137 @@
...
+                       |=================|      |
+                                |               | (recursive case)
+                  Transaction   |               |
+                                |---------------|
+                                |
+                                | (non recursive case)
+                                v
+                              finalTx : Transaction
+
+     */
```
dragospe, line 99: My mental model for this is basically "programming with arrows" (https://www.haskell.org/arrows/). Without much thought, I'm going to wager that it'd be a s
impler representation for these idea that what I currently have -- does scala have an equivalent?

src/main/scala/hydrozoa/infra/TxBuilder.scala
```diff
@@ -0,0 +1,137 @@
...
+
+        val cr: CalculationResult = this.calculate(qr) match {
+            case Left(err) => return Left(err)
+            case Right(r)  => r
+        }
+
+        val initTxBody: Transaction = this.txInitializer(cr, this.entryTx) match {
+            case Left(err) => return Left(err)
+            case Right(r)  => r
+        }
```
dragospe, line 117: Threading through error handling via "Either" is obnoxious. Using a monad -- or monad transformer/effects lib -- would probably simplify

src/main/scala/hydrozoa/l1/multisig/tx/initialization/InitializationTxBuilder.scala
```diff
@@ -1,21 +1,183 @@
...
+        Right(
+          entryTx.copy(
+            body = KeepRaw(
+              entryTx.body.value.copy(
+                inputs = Set(cr.seedUtxo),
+                outputs = IndexedSeq(cr.headTO, cr.changeTO).map(Sized(_)),
+                fee = feeCoin,
+                mint = Some(cr.beaconToken)
+              )
+            ),
```
dragospe, line 144: The TxInitializer gets a little thorny too. Optics would help here.

src/main/scala/hydrozoa/infra/TxBuilder.scala
```diff
@@ -0,0 +1,137 @@
...
+        finalTx
+    }
+    
+    
+    // Helper functions
+    // A function usable in the fixed-point that simply returns the initial transaction
+    def idFixedPoint(
+                        calculationResult: CalculationResult
+                    ): (rtx: Either[FixedPointError, Transaction], itx: Transaction) => Either[FixedPointError, Transaction] =
+        (r, i) => Right(i)
```
dragospe, line 136: N.B.: This doesn't seem to work correctly; it doesn't type check in the init Tx Builder and I'm not totally sure why.

src/main/scala/hydrozoa/l1/multisig/tx/initialization/InitializationTxBuilder.scala
```diff
@@ -1,21 +1,183 @@
...
+            isValid = true
+          )
+        )
+    }
+
+    override type FixedPointError = Void
+    override def txFixedPoint(calculationResult: CalculationResult): (
+        resultTxBody: Either[FixedPointError, Transaction],
+        initialTxBody: Transaction
+    ) => Either[FixedPointError, Transaction] = this.idFixedPoint
```
dragospe, line 155: (this is what doesn't type check)

src/main/scala/hydrozoa/infra/TxBuilder.scala
nau: Scala `Either` is right-biased like in Haskell so that should work
```scala
for 
  qr <- query(recipe)
  cr <- calculate(qr)
  ...
yield finalTx
```

src/main/scala/hydrozoa/l1/multisig/tx/initialization/InitializationTxBuilder.scala
```diff
@@ -1,21 +1,183 @@
...
+          )
+        )
+
+        // singleton beacon token minted by the native script with the TN being the hash of the
+        // seed utxo
+        // TODO: factor out "mkSingleToken"
+        val beaconToken: MultiAsset = Map(
+          (
+            headNativeScript.scriptHash,
+            Map(
```
nau, line 80:
```scala
Map(headNativeScript.scriptHash -> value)
// is the same as
Map( (headNativeScript.scriptHash, value) )

File: src/main/scala/hydrozoa/l1/multisig/tx/initialization/InitializationTxBuilder.scala (hunk 5 above)
Comment (nau) [reply src/main/scala/hydrozoa/l1/multisig/tx/initialization/InitializationTxBuilder.scala:155]: I don't think it's a valid Scala – you can't use parameter names in the return type. Try using multiple parameter lists (curried syntax):
```scala
def txFixedPoint(calculationResult: CalculationResult)(
        resultTxBody: Either[FixedPointError, Transaction],
        initialTxBody: Transaction
    ): Either[FixedPointError, Transaction]
```

src/main/scala/hydrozoa/infra/TxBuilder.scala
dragospe: Works perfect, thanks!

src/main/scala/hydrozoa/l1/multisig/tx/initialization/InitializationTxBuilder.scala
```diff
@@ -1,21 +1,183 @@
...
+    Transaction,
+    TransactionInput,
+    TransactionOutput,
+    TransactionWitnessSet,
+    Value
+}
+import scalus.builtin.Data.toData
+
+// FIXME: temporary value until we do proper fee calculation
+val feeCoin: Coin = Coin(1_000_000.toLong)
```
nau, line 31: 1_000_000L

PR #177, opened 2025-08-10
Type revision

dragospe:
Revise types to move away from bloxbean 

-----------
*Current status*: (forthcoming)


integration/src/test/scala/hydrozoa/model/HydrozoaState.scala
```diff
@@ -10,34 +10,36 @@ import hydrozoa.l2.ledger.L2Event
...
-    poolEvents: Seq[L2Event] = Seq.empty,
-
-    // L1
-    knownTxs: Map[TxId, TxL1] = Map.empty,
-    utxosActive: Map[UtxoIdL1, OutputL1],
-
-    // L2
-    utxosActiveL2: Map[UtxoIdL2, OutputL2] = Map.empty,
-    l2Tip: Block = zeroBlock
+                            peersNetworkPhase: PeersNetworkPhase,
```
euonymos, line 20: Is that how `scalafmtAll` formats that?

src/main/scala/hydrozoa/Types.scala
```diff
@@ -1,221 +1,317 @@
...
+
+type OutputNoTokens[L <: AnyLayer] = OutputNoTokens.OutputNoTokens[L]
+
+/* ---------------------------------------------------------------------------------------------
+ * UTxO
+ * Question: It seems that we go between using this type and just a raw tuple `(UtxoId[L], Output[L])`.
+ * If we use the tuple form, then we can ue an opaque type alias as above. Should we?
+ * ---------------------------------------------------------------------------------------------
+ */
+case class Utxo[L <: AnyLayer](input: UtxoId[L], output: Output[L])
```
euonymos, line 198: @GeorgeFlerovsky wanted to have an explicit _name_ for it, so I think we can use a type alias not a case class as @dragospe suggests.

src/main/scala/hydrozoa/l1/rulebased/onchain/UnliftMinimalValidator.scala
euonymos, line 1: We don't need this file anymore; it somehow survived in the string of merges.

src/main/scala/hydrozoa/l2/consensus/network/HeadPeerNetwork.scala
```diff
@@ -524,3 +521,29 @@ given JsonValueCodec[L2EventWithdrawal | L2EventTransaction] =
 
         override def nullValue: L2EventWithdrawal | L2EventTransaction = ???
     }
+
```
euonymos, line 524: Do we need those undefined instances, and if so, why?

src/main/scala/hydrozoa/node/rest/NodeRestApi.scala
```diff
@@ -1,15 +1,21 @@
 package hydrozoa.node.rest
 
+/* Question (Peter, 2025-08-10): My assumption is that this is the file is where we define our serialization
+boundary; i.e., where we go from an endpoint that contains a Bech32 string to a full-fledged ShelleyAddress.
+But this may be achievable earlier with tapirs schemas? I'm not sure.
+ */
+
```
euonymos, line 7: Yes, it was done here, but once Scalus has schema/json givens, we won't need it anymore.

PR #189, opened 2025-08-20
Outbox / Inbox actors

euonymos:


src/main/scala/hydrozoa/l2/consensus/network/outbox/DelivererActor.scala
```diff
@@ -0,0 +1,171 @@
...
+  *   - pulls missing messages from the database into [[queue]] using [[recover]] method
+  *   - subscribe to node's outbox actor once ready
+  *   - handle messages from the Outbox actor with [[enqueue]] method (and [[recover]] if needed)
+  *   - retry delivering to a particular peer till it succeeds
+  *
+  * TODO: update the following part
+  *
+  * Sending logic:
+  *
+  *   - Reads from the incoming channel, which connects to OutboxBroadcastActor.
```
dragospe, line 28: This particular line is out of date, right? We decided against using channels directly? (Disregarding that actors are using them underneath)

PR #200, opened 2025-09-10
Remove type tags, flesh out parsing function for deposit tx builder

dragospe:


src/main/scala/hydrozoa/multisig/ledger/l1/real/tx/InitializationTx.scala
```diff
@@ -1,25 +1,120 @@
...
 
 object InitializationTx {
     final case class Recipe(
-        headAddress: Address[L1],
-        fundingUtxos: List[Utxo[L1]],
-        fundingToTreasury: Coin
+        network: Network,
+        seedUtxo: (TransactionInput, TransactionOutput),
+        coins: BigInt,
+        peers: Set[VerificationKeyBytes]
```
dragospe, line 27: @GeorgeFlerovsky, please take a look at this. 

I've added the headaddress to the output of the Tx builder so that we don't have to do a pattern match, and revised the recipe to include the network and peer, which are needed to calculate the multisig script. Did you want to revise how we handle this?

GeorgeFlerovsky (approved):

src/main/scala/hydrozoa/multisig/ledger/l1/real/tx/SettlementTx.scala
```diff
@@ -13,21 +28,135 @@ final case class SettlementTx(
...
+        }
+
+    }
+
+    case class Recipe(
+        majorVersion: Int,
+        deposits: List[DepositUtxo],
+        utxosWithdrawn: Map[TransactionInput, TransactionOutput],
+        treasuryUtxo: TreasuryUtxo,
+        headNativeScript: Native,
```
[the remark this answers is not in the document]
GeorgeFlerovsky: Oh, good point.

src/main/scala/hydrozoa/multisig/ledger/l1/real/tx/InitializationTx.scala
GeorgeFlerovsky: I kinda want to memoize the script/address/policy when we initialize the LedgerL1 and just pass them to the tx builders

PR #208, opened 2025-09-15
Actor handles

GeorgeFlerovsky:
Good news: kinda succeeded (i.e. it compiles) implementing an abstract mechanism for defining:

- a synchronous request type (`RequestSync`) and its corresponding response type (`requestSync.Response`)
- an augmented type (specialized for actors) that includes the sync request and a deferred cell for the response (`ActorRequestSync`)
- methods to complete the deferred cell

Good news: it seems to work in `Persistence.scala`.

Bad news: I can't figure out how to work with a union type (A | B | C) and a type-level function mapped over the union (F(A) | F(B) | F(C)).

Seems too complicated, so I'm abandoning this PR for now.

### DIFF
--- a/src/main/scala/hydrozoa/lib/handle/Handle.scala
+++ b/src/main/scala/hydrozoa/lib/handle/Handle.scala
@@ -1,0 +1,21 @@
+package hydrozoa.lib.handle
+
+object Handle {
+    trait RequestAsync
+
+    trait RequestSync extends RequestAsync {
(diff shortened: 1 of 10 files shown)


GeorgeFlerovsky:
Perhaps the answer lies with intersection types somehow...

GeorgeFlerovsky:
Solved in a simpler way here: #209

PR #214, opened 2025-09-17
Convert to Scalus's TxBuilder and balancing routines

dragospe:


src/test/scala/hydrozoa/multisig/ledger/dapp/tx/DepositTx.scala
```diff
@@ -0,0 +1,103 @@
...
+                          throw RuntimeException(
+                            s"Produced deposit tx cannot be deserialized from CBOR"
+                          )
+                      case Right(cborParsed) if cborParsed != tx =>
+                          // println(ByteString.fromArray(tx.tx.toCbor).toHex)
+                          // assertEquals(expected = tx.tx.body.value.outputs(1), obtained = cborParsed.tx.body.value.outputs(1))
+                          assertEquals(tx, cborParsed)
+                      case _ => ()
+                  }
+          }
```
dragospe, line 100: @nau This where the tx gets built in the property test. The generator for the recipe  is above

nau: Fixed in latest Scalus

PR #219, opened 2025-09-26
Peter/tx builder 4

dragospe:


src/main/scala/hydrozoa/lib/tx/TxEditor.scala
```diff
@@ -177,6 +171,168 @@ object RedeemerManagement {
...
+  * transaction should fail with an insufficient fee error.
+  *
+  * If the number of expected signers is too large compared to the number of actual signers, the
+  * transaction may pass, but the fee will be over-paid.
+  *
+  * Note: this class is opaque. It is not (or should not be) possible to have expectedSigners that
+  * point to invalid transaction components
+  */
+
+case class TransactionWithSigners private (tx: Transaction, expectedSigners: Set[ExpectedSigner]):
```
dragospe, line 245: This should be the invariant of the case class: it shouldn't be possible to have a `txws : TransactionWithSigners` such that and `es` in `txws.expectedSigners` points to a transaction component that doesn't exist in `txws.tx`. 

I can write this more formally if it doesn't make sense.

src/main/scala/hydrozoa/lib/tx/TxBuilder.scala
dragospe, line 1: This file is a WIP. It does not compile currently, and contains some conflicting ideas.

src/main/scala/hydrozoa/lib/tx/TxBuilder.scala
```diff
@@ -50,47 +49,103 @@ instance Show TransactionBuilderStep where
   show = genericShow
  */
 
-sealed trait TransactionBuilderStep
+sealed trait TransactionBuilderStep:
+    val additionalSigners: Set[ExpectedSigner]
```
dragospe, line 53: my idea with this field is that each TBS should be associated with a set of ExpectedSigners.

For steps that are unlocked by PKHs, it should be a singleton set with just the `AddrKeyHash`. 

For steps that are locked by plutus or native scripts, it should be a user-provided set of `AddrKeyHash`s.

Then, when we apply the builder steps, we update the set of expected signers.  Additionally, if the builder step has a plutus script witness, we update the set of `RequiredSigners`.

src/main/scala/hydrozoa/lib/tx/TxEditor.scala
```diff
@@ -177,6 +171,168 @@ object RedeemerManagement {
...
+    ): Either[Set[ExpectedSigner], TransactionWithSigners] = {
+        val modifiedTx = f(this)
+        val signersPartition = TransactionWithSigners.partitionValidSignatures(
+          modifiedTx.tx,
+          modifiedTx.expectedSigners
+        )
+        if signersPartition.invalidSigners.isEmpty
+        then Right(new TransactionWithSigners(modifiedTx.tx, signersPartition.validSigners))
+        else Left(signersPartition.invalidSigners)
+    }
```
dragospe, line 278: Note that both of these currently only drop the signatures in the `_.expectedSigners` field. They should also drop signers from the `_.tx.body.value.requiredSigners`.

src/main/scala/hydrozoa/lib/tx/TxBuilder.scala
```diff
@@ -50,47 +49,103 @@ instance Show TransactionBuilderStep where
...
+    ) extends TransactionBuilderStep:
+        override val additionalSigners: Set[ExpectedSigner] = {
+            val signers = witness match {
+                case Some(ns: OutputWitness.NativeScriptOutput) => ns.additionalSigners
+                case Some(ps: OutputWitness.PlutusScriptOutput) => ps.additionalSigners
+                case None =>
+                    utxo.output.address match {
+                        case ShelleyAddress(_, payment: ShelleyPaymentPart.Key, _) =>
+                            Set(payment.hash)
+                        case _ => Set.empty // FIXME: What should we do in this case? Return a left?
```
dragospe, line 69: I'm pretty certain that `additionalSigners` should actually have type `Either[TxBuildError, Set[ExpectedSigner]]`. 

This particular condition is when a pkh utxo is passed, but the address is either a ShelleyAddress with a script component or a ByronAddress (are there other options?)

src/main/scala/hydrozoa/lib/tx/TxBuilder.scala
```diff
@@ -114,13 +169,24 @@ instance Show OutputWitness where
 
 sealed trait OutputWitness
```
dragospe, line 170: This should have `additionalSigners` as a method.

PR #239, opened 2025-10-11
Settlement/rollout transaction sequence (continued)

GeorgeFlerovsky:


src/main/scala/hydrozoa/multisig/ledger/dapp/txseq/tx/RolloutTx.scala
```diff
@@ -1,401 +1,513 @@
...
+        import State.Status
 
-        /** @param toWithdrawReversed
-          * @param mkSettlement
-          * @return
-          *   The settlement tx accompanied by a _chain_ of rollout transactions. TODO: make a
-          *   newtype
+        type Error = SomeBuildError
+
+        type Result = RolloutTx
```
dragospe: Can we get rid of these? I think think it makes the code less clear.

PR #250, opened 2025-10-18
Deinitialization tx

euonymos:


src/main/scala/hydrozoa/rulebased/ledger/dapp/tx/DeinitTx.scala
```diff
@@ -1,3 +1,216 @@
...
+                          )
+                      )
+                )
+
+            finalized <- context
+                .finalizeContext(
+                  protocolParams = env.protocolParams,
+                  diffHandler = new ChangeOutputDiffHandler(
+                    env.protocolParams,
+                    0 // the collateral sent back
```
GeorgeFlerovsky: Should this be a TODO comment or is `0` the correct argument here?

GeorgeFlerovsky (approved):
LGTM. Minor comments

PR #254, opened 2025-10-22
Peter/vendored scalus

dragospe:
more bumps as I fix other things upstream

### DIFF
--- a/src/main/scala/hydrozoa/Transitionary.scala
+++ b/src/main/scala/hydrozoa/Transitionary.scala
@@ -481,48 +481,7 @@ extension (self: TransactionOutput)
                     case None        => None
                 }
         }
-
-    /** Recursively calculate the minAda for UTxO using given parameters.
-      *
-      * @param params
-      *   Protocol params (for minAda calculation)
-      * @param update
-      *   A function that takes the calculated minAda for the [[candidateOutput]] and modifies the
-      *   output to calculate the new minAda. By default, it is [[replaceAdaUpdate]]
-      * @return
-      *   An output that has the [[update]] function applied to it until the minAda condition is
-      *   satisfied for the UTxO
-      */
-
-    def setMinAda(
-        params: ProtocolParams,
-        update: (Coin, TransactionOutput) => TransactionOutput = replaceAdaUpdate
-    ): TransactionOutput = {
-
-        @tailrec
-        def go(output: TransactionOutput): TransactionOutput =
-            val minAda = MinCoinSizedTransactionOutput(Sized(output), params)
-            if minAda <= output.value.coin
-            then output
-            else go(update(minAda, output))
-
-        go(self)
-    }
-
-/** A default update function for use with `setMinAda`. It replaces the output's coin with the given coin.
-  */
-def replaceAdaUpdate(coin: Coin, to: TransactionOutput): TransactionOutput = {
-    to match {
-        case s: TransactionOutput.Shelley => s.focus(_.value.coin).replace(coin)
-        case b: TransactionOutput.Babbage => b.focus(_.value.coin).replace(coin)
-    }
-}
-
-def keepRawL[A: Encoder](): Lens[KeepRaw[A], A] = {
-    val get: KeepRaw[A] => A = (kr => kr.value)
-    val replace: A => KeepRaw[A] => KeepRaw[A] = (a => kr => KeepRaw(a))
-    Lens[KeepRaw[A], A](get)(replace)
-}
+    def ensureMinAda(params: ProtocolParams) : TransactionOutput = TransactionBuilder.ensureMinAda(self, params)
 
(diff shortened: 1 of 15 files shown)


src/main/scala/hydrozoa/Transitionary.scala
```diff
@@ -479,48 +481,7 @@ extension (self: TransactionOutput)
...
-        case b: TransactionOutput.Babbage => b.focus(_.value.coin).replace(coin)
-    }
-}
-
-def keepRawL[A: Encoder](): Lens[KeepRaw[A], A] = {
-    val get: KeepRaw[A] => A = (kr => kr.value)
-    val replace: A => KeepRaw[A] => KeepRaw[A] = (a => kr => KeepRaw(a))
-    Lens[KeepRaw[A], A](get)(replace)
-}
+    def ensureMinAda(params: ProtocolParams) : TransactionOutput = TransactionBuilder.ensureMinAda(self, params)
```
dragospe: this now uses the upstream version

src/main/scala/hydrozoa/multisig/ledger/dapp/tx/Tx.scala
```diff
@@ -43,7 +46,38 @@ object Tx {
...
+            def explainModify(modifyString : String => String)(implicit line: Line, file: File, enclosing: Enclosing):
+                Either[(E, String), A] = {
+                augmentedEither.left.map(t => (t._1, modifyString(t._2)))
+            }
+
+            def explainAppendConst(string : String)(implicit line: Line, file: File, enclosing: Enclosing):
+                Either[(E, String), A] = {
+                augmentedEither.explainModify(_ + s";\n[${file.value}:${line.value} in ${enclosing.value}] $string")
+            }
+
```
dragospe: This was a quick hack-in because I needed more information in the errors.

What I'd like to do properly is place these in a monad that gives us:

- Actual stack traces, such that we get a proper `List[Location]`, where each subsequent re-throw appends their stack trace
- A way to associate stack traces with individual messages, so that each `Location` can add more information in a structured way
- Probably actual typed errors 

The idea would be to have something like 

```
type Error[E,A] = Either[(E, Seq[(Location, Message)]), A]

val foo : Error[FooError, Bar] = 
{

  for { 
    bar <- Right(123) 
    baz <- tryBaz.explainAppend("foo threw an error on tryBaz")
  
  } yield baz
}
```

where we're get something like 

```
Left(Vector(Foo, 
  ("/home/(...)/Foo.scala", "foo threw an error on tryBaz"),
  ("/home/(...)/Baz.scala", "tryBaz failed because Qux was negative. Qux = -1"),
  ("/home/(...)/Qux.scala", "Qux could not reach server. Return code: -1")
  )
)
```

I wonder if there is a library that already does this, or if we could modify a logging library?

src/test/scala/hydrozoa/multisig/ledger/dapp/tx/SettlementTx.scala
```diff
@@ -136,23 +137,32 @@ def genSettlementTxSeqBuilder(
...
     network: Network = testNetwork
 ): Gen[(SettlementTxSeq.Builder, SettlementTxSeq.Builder.Args, NonEmptyList[TestPeer])] = {
+    // A helper to generator empty, small, medium, large (up to 1000)
+    def genHelper[T](gen : Gen[T]) : Gen[Vector[T]] = Gen.sized(size =>
+      Gen.frequency(
+       (1, Gen.const(Vector.empty)),
+       (2, Other.vectorOfN(size, gen)),
+       (5, Other.vectorOfN(size * 5, gen)),
+        (1, Other.vectorOfN(1, gen))
+      ).map(_.take(1000))
```
dragospe: I need to learn how to dump the actual generated sizes, because I want to observe statistics on the actual data generated. I would have saved myself a lot of time if I realized that `Coin` was previously generating very small values.

dragospe: Alright, after playing around with `scalatest`, `munit`, and `scalacheck` directly, I'm thinking that maybe we should just be using `scalacheck` by itself. 

If we want to stick with `scalatest`, we lose the ability to print out test statistics on successful results as far as I can tell. With scalacheck, we can get

```
+ SettlementTxSeq Builder Properties.Build settlement tx sequence: OK, pass
  ed 100 tests.
> Collected test data: 
38% with rollouts, 100 < numDeposits < 500
21% no rollouts, numDeposits <= 100
20% with rollouts, numDeposits <= 100
11% with rollouts, 500 < numDeposits <= 1000
8% no rollouts, 100 < numDeposits < 500
2% no rollouts, 500 < numDeposits <= 1000
```

The alternative is either to wing it and hope our generators work, or to test them directly

PR #256, opened 2025-10-26
Coin arithmetic draft

euonymos:


src/main/scala/hydrozoa/lib/coin/Coin/Coin.scala
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    // ===================================
+    // Coin.Fractional
+    // ===================================
+
+    type Fractional = Fractional.Type
+
+    object Fractional {
+        opaque type Type = BigDecimal
```
dragospe: Why not rational?

dragospe (commented):
Going to leave this here:

https://github.com/NorfairKing/really-safe-money

there used to be a good blog post accompanying the original `safe-moeny` library, but it seems that the original author have made a bunch of repos and that post private. 

Here's a different post linked in that repo: 
https://cs-syd.eu/posts/2022-08-22-how-to-deal-with-money-in-software
I haven't read it closely (and its been a while since I read the original safe-money post) but we (and scalus) can probably take lessons from there

PR #268, opened 2025-12-01
Peter/metadata and parsing updates

dragospe:


src/main/scala/hydrozoa/multisig/ledger/dapp/tx/Metadata.scala
```diff
@@ -1,111 +1,233 @@
...
-                    Map(
-                      Metadatum.Text(typeToString(txType))
-                          -> Metadatum.Bytes(headAddress.toBytes)
-                    )
+
+    given Encoder[ShelleyAddress] =
+      summon[Encoder[Address]].asInstanceOf[Encoder[ShelleyAddress]]
+
+    given Decoder[ShelleyAddress] =
+      summon[Decoder[Address]].asInstanceOf[Decoder[ShelleyAddress]]
```
dragospe: There is an encoder for `Address`, but not directly an encoder for `ShelleyAddress`. I asked if there was a better way to do this, and I'll update if I hear a response.

PR #292, opened 2025-12-23
Add `TestM`

dragospe:
An initial implementation of `TestM`. The API can be expanded with additional lifts as necessary.

Pay particular attention to the `Requests` and `Scenarios` objects within `JointLedgerTest`


src/test/scala/hydrozoa/multisig/ledger/DappLedgerTest.scala
dragospe: @euonymos : Don't focus too much on this one, I didn't update it much because I'll be refactoring DappLedger significantly this week

PR #301, opened 2025-12-31
Peter/joint ledger updates

dragospe:


src/main/scala/hydrozoa/multisig/consensus/CardanoLiaison.scala
```diff
@@ -505,7 +509,7 @@ trait CardanoLiaison(config: Config) extends Actor[IO, Request] {
                               fallbackValidityStart
                             ) match {
                                 // (1)
-                                case _ if currentSlot < happyPathTxTtl =>
+                                case _ if currentSlot < happyPathTxTtl.toSlot(config.slotConfig) =>
```
GeorgeFlerovsky: CardanoLiaison shouldn't use the "current slot" from Blockfrost.
Instead, it should use the current wall-clock time from `IO.realTime`.

Corrolary: CardanoBackend's `GetHeadState.Resp` should not include the current slot.

GeorgeFlerovsky (commented):
Reviewed part of the PR. I'll continue later

src/main/scala/hydrozoa/multisig/consensus/CardanoLiaison.scala
GeorgeFlerovsky: Created issue: #304

PR #305, opened 2026-01-06
Consensus actor

euonymos:
- [x] Split up the own acks queue  by purring postponed acks into cells,
- [x] Use Scalus builtin types/functions for: verification
- [x] Use Scalus builtin types/functions for: adding a signature
- [x] Use VerifiedSignaturesInWitnessesValidator to implement the verification function
- [x] Add finalization flag to acks
- [x] Add block confirmed request type to CardanoLiaison, EventSequencer, BlockWeaver, and PeerLiaison. Make ConsensusActor send these requests when appropriate.
- [x] Major/final cell completion
- [x] Review state updates
- [x] Send final block to Peer Liaison
- [x] Remove TxSeq types from Cardano Liaison
- [x] Move parts from ConsensusProtocol to actors
- [x] Remove persistence
- [x] Check tests

### DIFF
--- a/src/main/scala/hydrozoa/multisig/ledger/dapp/tx/Tx.scala
+++ b/src/main/scala/hydrozoa/multisig/ledger/dapp/tx/Tx.scala
@@ -21,6 +21,13 @@ trait Tx[Self <: Tx[Self]] { self: Self =>
       * Unfortunately this can't be generalized since Focus requires a concrete type.
       */
     def txLens: Lens[Self, Transaction]
+
+    /** This excludes the lens from equality. */
+    override def equals(obj: Any): Boolean = obj match {
+        case that: Tx[?] =>
+            this.tx == that.tx
+        case _ => false
+    }
 }
 
 trait HasValidityStart:
--- a/src/main/scala/hydrozoa/multisig/consensus/BlockWeaver.scala
+++ b/src/main/scala/hydrozoa/multisig/consensus/BlockWeaver.scala
@@ -14,12 +14,13 @@ import hydrozoa.multisig.protocol.types.Block.blockEvents
 import hydrozoa.multisig.protocol.types.{Block, LedgerEvent, LedgerEventId, Peer}
 
 /** Block weaver actor.
-  *   - When the node is leading a block, packages known unprocessed and incoming events, i.e., L1
-  *     deposits and L2 txs into a new block.
-  *   - When follower, before any block arrives, simply store incoming events in the mempool,
-  *     keeping their order or arrival.
-  *   - When a block arrives, feeds all the block to the joint ledger or uses Awaiting mode to wait
-  *     till all events get through.
+  *   - When the node is leading a block, the weaver packages all known unprocessed (by the time
+  *     block is stared) and continue streaming all incoming events (until the block is completed)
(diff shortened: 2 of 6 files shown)


src/main/scala/hydrozoa/multisig/consensus/ConsensusActor.scala
```diff
@@ -0,0 +1,498 @@
...
+                state.withConsensusFor(minor)(_.applyAugBlock(minor))
+            case major: AugmentedBlock.Major =>
+                state.withConsensusFor(major)(_.applyAugBlock(major))
+            case final_ : AugmentedBlock.Final =>
+                state.withConsensusFor(final_)(_.applyAugBlock(final_))
+        }
+    } yield ()
+
+    def handleAck(
+        ack: AckBlock.Minor | AckBlock.Major1 | AckBlock.Major2 | AckBlock.Final1 | AckBlock.Final2
```
GeorgeFlerovsky: Why not just `ack: AckBlock`?

src/main/scala/hydrozoa/multisig/consensus/ConsensusActor.scala
```diff
@@ -0,0 +1,498 @@
...
+    object BlockConfirmed:
+
+        final case class Minor(
+            augBlock: AugmentedBlock.Minor,
+            headerSignatures: Set[ByteString],
+            // Fully signed txs
+            immediateRefundsSigned: List[RefundTx.Immediate],
+            // Fully signed txs
+            postDatedRefundsSigned: List[RefundTx.PostDated]
+        ) extends BlockConfirmed
```
GeorgeFlerovsky: I thought the plan was to have different `BlockConfirmed` message types, each with different information provided as relevant to its recipient (BlockWeaver, CardanoLiaison, EventSequencer, PeerLiaison).

src/main/scala/hydrozoa/Types.scala
```diff
@@ -263,6 +243,7 @@ extension (utxo: UTxO)
 // Keys
 
 // A verification key of a peer, used on both L1 and L2
+// TODO: review, I think it's strange. Shall we use opaque type instead?
 case class VerificationKeyBytes(bytes: ByteString) {
```
dragospe: I agree. What constructors would we need?

The valid ways to generate one would be...

- Key generation
- (unsafe) key parsing

anything else?

src/main/scala/hydrozoa/multisig/ledger/dapp/tx/Tx.scala
```diff
@@ -12,8 +13,14 @@ import scalus.cardano.txbuilder.TransactionBuilder.ResolvedUtxos
 import scalus.cardano.txbuilder.{ChangeOutputDiffHandler, Environment, SomeBuildError, TransactionBuilder}
 import sourcecode.*
 
-trait Tx {
+trait Tx[Self <: Tx[Self]] {
```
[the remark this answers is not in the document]
dragospe: Oh, I just looked above.

So now instead of `case class FooTx extends Tx`, we need `case class FooTx extends Tx[FooTx]`? 

This is kinda a "trait class" or something, conceptually.

Issue #296, opened 2025-12-29
Hydrozoa L1 tx timing


GeorgeFlerovsky:
## Initialization and initial fallback

Parameters:
- `system_start_time` = some absolute point in time (POSIX)
- `min_settlement_duration` = 12 hours
- `inactivity_margin_duration` = 24 hours
- `silence_duration` = 5 minutes

We consider a hydrozoa head to be started as of its `system_start_time`, defined in its configuration as some arbitrary point in the past.

The initialization tx must not have a validity start time, and its validity end time must be:
```
fallback(0).validity_start = ∅
initialization.validity_end = system_start_time + min_settlement_duration + inactivity_margin_duration
```

The initial fallback tx (i.e. Fallback 0 from initialization) must not have a validity end time, and its validity start time must be:
```
fallback(0).validity_start = initialization.ttl + silence_period
fallback(0).validity_end = ∅
```

## Subsequent settlement and fallback

Settlement and fallback transactions have validity start and end times defined, in part, relative to their predecessors:
```
settlement(N+1).validity_end = fallback(N).validity_start - silence_period
settlement(N+1).validity_start = ∅

fallback(N+1).validity_start =
  settlement(N+1).validity_start + min_settlement_duration + inactivity_margin_duration + silience_period
fallback(N+1).validity_end = ∅
```

...

dragospe:
@GeorgeFlerovsky , `TxTiming` currently has a `initializationFallbackDeviation` with a default time of 10 minutes. It has the following comment:


```scala
  * @param initializationFallbackDeviation
  *   Since it doesn't make sense to ask users to specify exact the same TTL/validity start slot for
  *   txs in the initialization sequence that we may calculate based on the time the initialization
  *   request was received, we need to allow some deviation which is defined by that parameter. The
  *   rule is that the specified value in the txs should stay in the [calculatedTime -
  *   initializationFallbackDeviation; calculatedTime + initializationFallbackDeviation].
```

and used in the initialization tx as follows:

```scala
   possibleRange = (
              toSlot(
                (preciseFallbackValidityStart - txTiming.initializationFallbackDeviation).toMillis
              ),
              toSlot(
                (preciseFallbackValidityStart + txTiming.initializationFallbackDeviation).toMillis
              )
            )

            _ <-
                if fallbackValidityStartSlot < possibleRange._1 || fallbackValidityStartSlot > possibleRange._2
                then
                    Left(
                      FallbackTxValidityStartError(
[...]

Issue #304, opened 2026-01-05
CardanoLiaison shouldn't use the "current slot" from Blockfrost.


GeorgeFlerovsky:
CardanoLiaison shouldn't use the "current slot" from Blockfrost.
Instead, it should use the current wall-clock time from `IO.realTime`.

Corrolary: CardanoBackend's `GetHeadState.Resp` should not include the current slot.

_Originally posted by @GeorgeFlerovsky in https://github.com/cardano-hydrozoa/hydrozoa/pull/301#discussion_r2663023173_

PR #306, opened 2026-01-06
Joint Ledger Updates

dragospe:
This PR primarily contains deposit timing related functionality

- pass initialFallbackValidityStart to JointLedger
- add `QuantizedTime` object and related types
- add `TestControl` runner to `PropertyM` (althought we don't actually use it right now)
- add slotConfig to block weaver (to allow for quantizing the call to IO.realTime)
- Ensure that deposit that have a validity end prior to the block start time never enter into the deposit
- Change the `Boolean` in `events: List[(LedgerEventId, Boolean)]` to `ValidityFlag`
- Revise `depositPartition` to split into three partitions, according to #296 \
- Make `RefundInstructions` opaque and ensure that the refund start time is derived from a `QuantizedInstant`
- And assertions to both builder and parser for deposit Tx that the following three times are coherent according to #296:
  - The deposit tx validity end
  - the refund start time, as carried in the deposit utxo datum
  - the refund start time, as represented in the refund tx body
- Change TxTiming to use quantized times.
- Fix a bug when parsing refund txs where the witness sets would not match between the parsed value and the constructed value
- Add the fallbackTx to the BlockEffects.Initial
- Expand the arguments represented in the RegisterDepositRequest
- Change usages of `IO.monotonic` to `IO.realTimeInstant`
- Revise tests to use quanitzed timing
- Add GenesisObligation CBOR roundtripping test  
- Make TestM Polymorphic, and add a specialized variant to the joint ledger test
- Add Deposit timing tests
- Make initialization tx parser fail if timing is invalid

### DIFF
--- a/src/main/scala/hydrozoa/multisig/ledger/DappLedgerM.scala
+++ b/src/main/scala/hydrozoa/multisig/ledger/DappLedgerM.scala
@@ -80,8 +80,6 @@ object DappLedgerM {
       * NOTE: This checks SOME time bounds. Specifically, it checks whether the deposit's absorption
       * validity period ended prior to the start of the current block.
       */
-    // TODO: Return the produced deposit utxo and a post-dated refund transaction for it.
-    // Note from Peter (2026-01-09): I think the above TODO is stale?
     def registerDeposit(req: RegisterDeposit): DappLedgerM[Unit] = {
         import req.*
         for {
@@ -89,7 +87,7 @@ object DappLedgerM {
             parseRes =
                 DepositRefundTxSeq
                     .parse(
-                      depositTxBytes = serializedDeposit,
+                      depositTxBytes = depositTxBytes,
                       refundTxBytes = refundTxBytes,
                       config = config,
                       virtualOutputsBytes = virtualOutputsBytes,
--- a/src/main/scala/hydrozoa/multisig/protocol/types/LedgerEvent.scala
+++ b/src/main/scala/hydrozoa/multisig/protocol/types/LedgerEvent.scala
@@ -66,7 +66,7 @@ object LedgerEvent {
     ) extends LedgerEvent
 
     final case class RegisterDeposit(
-        serializedDeposit: Array[Byte],
+        depositTxBytes: Array[Byte],
         refundTxBytes: Array[Byte],
         eventId: LedgerEventId,
         virtualOutputsBytes: Array[Byte],
--- a/src/test/scala/hydrozoa/multisig/ledger/JointLedgerTest.scala
+++ b/src/test/scala/hydrozoa/multisig/ledger/JointLedgerTest.scala
@@ -394,7 +394,7 @@ object JointLedgerTestHelpers {
 
                 req =
                     RegisterDeposit(
-                      serializedDeposit = signTx(peer, depositRefundTxSeq.depositTx.tx).toCbor,
+                      depositTxBytes = signTx(peer, depositRefundTxSeq.depositTx.tx).toCbor,
                       refundTxBytes = signTx(peer, depositRefundTxSeq.refundTx.tx).toCbor,
                       donationToTreasury = Coin.zero,
                       virtualOutputsBytes = virtualOutputsBytes,


src/main/scala/hydrozoa/lib/cardano/scalus/QuantizedTime.scala
```diff
@@ -0,0 +1,239 @@
...
+     technically.
+  - It should be possible to pass [SC <: SlotConfig] as a type-level literal in the parameters. But I tried it briefly,
+    and couldn't quite figure out how to make things unify correctly.
+    - This would potentially move the runtime [[require]] to a compile time check.
+  - These could probably also be made sub-types of Instant/FiniteDuration, and potentially opaque type aliases?
+  - Ensure soundnes of algebraic operations when combining quantized and unquantized values
+    - i.e., is quantizedInstant + finiteDuration.quantize(quantizedInstant.slotConfig)
+      always equal to QuantizedInstant(quantizedInstant.instant + finiteDuration, quantizedInstant.slotConfig) ?
+ */
+object QuantizedTime {
```
euonymos: This makes sense, but how is it better than just slots, or _fancy slots_ I should say? Slots are _precisely_ quantized time.

src/main/scala/hydrozoa/lib/cardano/scalus/QuantizedTime.scala
```diff
@@ -0,0 +1,239 @@
...
+
+    case class QuantizedInstant private (instant: java.time.Instant, slotConfig: SlotConfig) {
```
euonymos: I see why we need `slotConfig`, but shouldn't it be just a global constant? Do we really need to support `QuantizedInstant`s with different slot configs?

src/main/scala/hydrozoa/multisig/consensus/CardanoLiaison.scala
```diff
@@ -304,9 +303,7 @@ class CardanoLiaison(config: CardanoLiaison.Config, stateRef: Ref[IO, CardanoLia
                         then IO.pure(dueActions)
                         else {
                             lazy val initAction = {
-                                if l1State.currentSlot < config.initializationTx.validityEnd.toSlot(
-                                      config.slotConfig
-                                    )
+                                if l1State.currentSlot < config.initializationTx.validityEnd.toSlot
```
euonymos: This is going to change - @GeorgeFlerovsky told me we don't want to use Cardano L1 time (slots) at all. So it should be `now <- IO.realTime.map(_.toEpochQuantizedInstant(config.slotConfig))` I guess.

GeorgeFlerovsky: Actually, we shouldn't be getting the current slot from Cardano Backend, at all.
We should get the wall-clock time from `IO.realTime` (perhaps quantizing it).

I mentioned this in a previous review https://github.com/cardano-hydrozoa/hydrozoa/pull/301#discussion_r2663023173 and created an issue #304.

src/main/scala/hydrozoa/multisig/ledger/DappLedgerM.scala
```diff
@@ -73,28 +74,40 @@ object DappLedgerM {
...
+      * NOTE: This checks SOME time bounds. Specifically, it checks whether the deposit's absorption
+      * validity period ended prior to the start of the current block.
       */
     // TODO: Return the produced deposit utxo and a post-dated refund transaction for it.
-    def registerDeposit(
-        serializedDeposit: Array[Byte],
-        eventId: LedgerEventId,
-        virtualOutputs: NonEmptyList[GenesisObligation]
-    ): DappLedgerM[Unit] = {
+    // Note from Peter (2026-01-09): I think the above TODO is stale?
```
GeorgeFlerovsky: Yep, the above comment is stale. The Dapp Ledger no longer originates the post-dated refund transaction; it merely verifies the user-provided one (by reproducing it).

src/main/scala/hydrozoa/multisig/protocol/types/LedgerEvent.scala
```diff
@@ -57,13 +65,14 @@ object LedgerEvent {
...
     ) extends LedgerEvent
 
-    // FIXME: This should include the refundTxBytes
-    // FIXME: The virtual outputs should not be parsed yet (i.e. Array[Byte])
     final case class RegisterDeposit(
-        override val eventId: LedgerEventId,
         serializedDeposit: Array[Byte],
-        // TODO: Pop up [[GenesisObligation]]?
-        virtualOutputs: NonEmptyList[GenesisObligation]
+        refundTxBytes: Array[Byte],
```
GeorgeFlerovsky: Since we're calling this `refundTxBytes`, I guess we should also rename `serializedDeposit` into `depositTxBytes`.

src/main/scala/hydrozoa/multisig/ledger/JointLedger.scala
```diff
@@ -408,14 +411,13 @@ final case class JointLedger(
 
             // For a description of timing, see GitHub issue #296
             forceUpgradeToMajor: Boolean =
-                if producing.competingFallbackValidityStart.isEmpty
-                then true
-                else
-                    txTiming.minSettlementDuration.toMillis >=
+                // First block is always major
```
GeorgeFlerovsky: Why do you claim that the first block (i.e. Block 1, immediately following the initial block) is always major? IMO it can be minor

src/main/scala/hydrozoa/lib/cardano/scalus/QuantizedTime.scala
dragospe: My first attempt was actually to carry this around as a type parameter so that the `require`s checks were not necessary. It looked something like

```scala
QuantizedInstant[SC <: SlotConfig](instant : java.time.Instant)
```

using a type-literal SlotConfig. Hypothetically, it shouldn't be much more complex that type-safe `Vector[N <: Nat]`, but I wasn't totally sure how to coax the type system into not throwing inscrutable errors.

Maybe a future exploration if we have the time. I wonder if it could be made approachable/maintainable.

src/main/scala/hydrozoa/multisig/ledger/JointLedger.scala
```diff
@@ -367,16 +366,20 @@ final case class JointLedger(
                   )
                 )((acc, deposit) =>
                     val depositValidityEnd =
-                        java.time.Instant.ofEpochMilli(
-                          deposit._2.datum.refundInstructions.startTime.toLong
-                        ) - txTiming.silenceDuration
-                    val depositAbsorptionStart: java.time.Instant =
+                        deposit._2.datum.refundInstructions.startTime
```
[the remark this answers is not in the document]
dragospe: https://github.com/cardano-hydrozoa/hydrozoa/issues/309

PR #307, opened 2026-01-13
Cardano backend

euonymos:
1. Remove Bloxbean. It's in the deps since Scalus transitively brings it up. We use only:
    1. HexUtil
    2. Cardano backend Blockfrost (only backend-client)
    3. Yaci backend (tests only)
2. Cardano backend + Yaci integration:
    * Turn Cardano backend actor / Cardano backend protocol into a CardanoBackend[F[_]] with the following implementations:
        * L1 mock
        * Blockfrost (for Yaci and public networks)
    * Update Cardano Liaison
        * Stop using sync ?: - instead, the liaison owns and use CardanoBackend[IO] instance
        * Pass the head state to the block weaver
3. Temporarily, the Cardano Liaison test suite is disabled, since now we need an instrumented version of IO.realTime to test rollbacks (previously, the depth of rollback was obtained as a slot number from the backend, likely in the wrong manner).
4. Some improvements on QuantizedTime from @GeorgeFlerovsky

### DIFF
--- a/src/test/scala/hydrozoa/time/TimeActor.scala
+++ b/src/test/scala/hydrozoa/time/TimeActor.scala
@@ -1,0 +1,5 @@
+package hydrozoa.time
+
+class TimeActor {
+
+}
--- a/src/test/scala/hydrozoa/time/TimeActorSpec.scala
+++ b/src/test/scala/hydrozoa/time/TimeActorSpec.scala
@@ -1,0 +1,5 @@
+package hydrozoa.time
+
+class TimeActorSpec {
+
+}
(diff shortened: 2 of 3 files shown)


src/test/scala/hydrozoa/time/TimeActorSpec.scala
dragospe: This can probably be removed?

src/test/scala/hydrozoa/time/TimeActor.scala
dragospe: This can probably be removed as well

PR #313, opened 2026-01-16
Update peer liaison

GeorgeFlerovsky:
Features/fixes:
- [x] Fix peer liaison connections to block weaver and consensus actor.
- [x] Update and migrate peer liaison request types from ConsensusProtocol to PeerLiaison
- [x] Review peer liaison code and add TODOs for things that need to be fixed/updated/implemented.
- [x] Fix peer liaison's append function for outbox queues -- instead of just incrementing by one, look at the block/ack/event Id and check if it's incremented as expected.
- [x] Add field to state: the current GetMsgBatch that the peer liaison is sending to its counterpart.
- [x] Compare any received batch from counterpart against the current GetMsgBatch in state
  - [x] Items in batch must immediately follow the indices in GetMsgBatch. (For blocks, "immediately follows" means next block number for which the counterpart wil be leader)
  - [x] There must be no gaps in the list of events.
- [x] Make the consensus actor send confirmed blocks to all peer liaisons, so that they can remove confirmed blocks, acks, and events from their outbox queues.
- [x] Make the consensus actor send confirmed blocks to the event sequencer, including the signed post-dated refund txs.

General clean-up:
- [x] Eliminate `ConsensusProtocol.scala`
- [x] Eliminate `LedgerProtocol.scala`
- [x] Eliminate `ManagerProtocol.scala`

### DIFF
--- a/src/main/scala/hydrozoa/multisig/consensus/ConsensusActor.scala
+++ b/src/main/scala/hydrozoa/multisig/consensus/ConsensusActor.scala
@@ -4,6 +4,7 @@ import cats.effect.{IO, Ref}
 import cats.implicits.*
 import com.suprnation.actor.Actor.{Actor, Receive}
 import com.suprnation.actor.ActorRef.ActorRef
+import com.suprnation.typelevel.actors.syntax.BroadcastOps
 import hydrozoa.multisig.ledger.dapp.tx.{DeinitTx, FallbackTx, FinalizationTx, RefundTx, RolloutTx, SettlementTx, Tx}
 import hydrozoa.multisig.protocol.types.AckBlock.HeaderSignature.given
 import hydrozoa.multisig.protocol.types.AckBlock.{HeaderSignature, TxSignature}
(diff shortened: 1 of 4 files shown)


euonymos (approved):

src/main/scala/hydrozoa/multisig/protocol/types/LedgerEvent.scala
```diff
@@ -36,6 +36,10 @@ object LedgerEventId {
         def peerNum: Peer.Number = Peer.Number(self._1)
         def eventNum: Number = Number(self._2)
 
+        def precedes(other: Id): Boolean =
+            self.peerNum == other.peerNum &&
+                self.eventNum.increment == other.eventNum
```
dragospe: perhaps 

```suggestion
       /**
       * Returns `true` when given a `LedgerEvent.Id` that is from the same peer and immediately precedes this event in terms of event numbering
       */
        def immediatelyPrecedes(other: Id): Boolean =
            self.peerNum == other.peerNum &&
                self.eventNum.increment == other.eventNum
```
?

PR #330, opened 2026-01-27
Integration tests are back: stage 1

euonymos:
- [x] Fix justfile/build.sbt to lint/fmt subprojects (didn't work previously)
- [x] Raw config generation
- [x] Parse config 
- [x] Fix bug: consensus actor major/final cells should await own ack2 before switching to stage two
- [x] Rename "payouts" to "disbursements" in multisig Deinit
- [x] Agent actor for stage 1
- [x] Conditions for block briefs, CompleteBlock command
- [x] Move the god IO action into TestControl (should be optional though, since we don't need it for Yaci)
- [x] TestControl-based outer driver
- [x] Check: backend is aware of all txs that should be submitted according to the model state
- [x] Run the actors and the actor system
- [x] Stage 1: joint ledger + Cardano liaison integration happy-path with L1 mock
- [x] Improve output - use logging
- [x] Add logging
- [x] Fix block promotion property failure
- [x] Fix "last executed command" -> "failed command", fix the number
- [x] Automate command numbering
- [x] Reorganize Stage 1 suite
- [x] Count simulated time
- [x] Use ByteString for KzgCommitment
- [x] Move out `HeadConfig` from the Stage1State: only things really needed should be left

Leaving for the following PRs:
- [ ] Stage 1: joint ledger + Cardano liaison integration happy-path with Yaci
- [ ] Stage 1: joint ledger + Cardano liaison integration happy-path with public testnet
- [ ] Joint ledger should panic if block creation time falls on or after the silence period of the competing fallback
- [ ] Add deposits
- [ ] More diverse and realistic events
- [ ] Remove all defaults like: network: Network = testNetwork
- [ ] Check all addresses belong to the same network upon parsing HeadConfig

### DIFF
--- a/integration/src/test/resources/logback.xml
+++ b/integration/src/test/resources/logback.xml
@@ -9,8 +9,8 @@
         <appender-ref ref="STDOUT"/>
     </root>
     <logger name="org.scalacheck.commands.ModelBasedSuite" level="warn"/>
-    <logger name="hydrozoa.integration.stage1.Suite" level="warn"/>
     <logger name="Stage1.Commands" level="warn"/>
+    <logger name="Stage1.Suite" level="warn"/>
     <logger name="CardanoLiaison" level="warn"/>
     <logger name="scalus" level="warn"/>
 </configuration>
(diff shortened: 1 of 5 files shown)


dragospe (approved):
Overall, it seems reasonable. 

I'd need some time to work with it before giving more detailed feedback, but I'm pretty confident I could figure it out.

One thing you mentioned -- you said that some of this was vendored from `scalacheck`. Do you need to put the [license file](https://github.com/typelevel/scalacheck/blob/main/LICENSE) anywhere?

It may be worthwhile to document the differences between this and the upstream as well, in lieu of writing a full HOWTO.

PR #419, opened 2026-04-08
Feat: export onchain scripts using CIP-57

euonymos:
# CIP-57 Blueprint Script Export

* Stores the blueprints in `main/resources/plutus.json`
* There is a new just target to export: `just export-scripts`
* There is a test suite that compiles scripts and verifies lots of stuff:
    * All scripts are present
    * Hashes and compiled code are the same
    * Titles are correct - this is important since titles are used as identifiers
    * All datums/redeemers are defined
* Removed almost all dirty stuff from scripts' companion objects
* Added `HydrozoaBlueprint` that allows loading blueprints
* RB regime actors and `ScriptReferenceUtxo` from the config now use it not the companion objects
* All tests now:
    * Don't contruct the address manually - we tender to ignore `address` method
    * Use pre-compiled script to get the address, no compilation is needed, faster
* Removed unnecessary size/hash tests from the old scripts' test suits


src/main/resources/plutus.json
```diff
@@ -0,0 +1,195 @@
+{
+  "preamble": {
+    "title": "Hydrozoa Rule-Based Regime Validators",
+    "description": "CIP-57 compliant blueprint for Hydrozoa's rule-based state channel validators. These validators manage dispute resolution and treasury operations when multi-party consensus is no longer held.",
+    "version": "1.0.0",
+    "compiler": {
+      "name": "scalus",
+      "version": "0.15.1"
```
dragospe, line 8: should we also include a git commit?

src/main/scala/hydrozoa/config/HydrozoaBlueprint.scala
```diff
@@ -0,0 +1,151 @@
...
+                )
+            }
+            val source = scala.io.Source.fromInputStream(stream)
+            try {
+                val json = source.mkString
+                Blueprint.fromJson(json)
+            } finally {
+                source.close()
+                stream.close()
+            }
```
dragospe: You can also use `Resource` from `cats.effect`, if you'd like.

src/test/scala/hydrozoa/rulebased/ledger/l1/script/plutus/DisputeResolutionScriptTest.scala
dragospe, line 1: Delete?

dragospe:
So if I'm a user and I want to make sure that a blueprint matches the actual compiled scripts, I should:

- Checkout the appropriate version of hydrozoa
- Export the blueprint
- Verify it against the blueprint I have

Right?

dragospe:
I think we should add a flag to the export utilities that strips out debugging messages from the scripts for production. 
We don't use many right now in the happy-path, but it is extremely useful to have more verbose output when doing more exhaustive testing of onchain code. This could be good to add while the code is fresh in your mind.

What do you think?

src/main/resources/plutus.json
euonymos: Would be nice, but CIP-57 doesn't seem to define such a field.

src/test/scala/hydrozoa/rulebased/ledger/l1/script/plutus/DisputeResolutionScriptTest.scala
euonymos: I left it so we can have a place to add unit tests for the script in the future if we want to.

euonymos:
@dragospe:

> So if I'm a user and I want to make sure that a blueprint matches the actual compiled scripts, I should:

As an end-user, you just use pre-deployed versions of scripts, which are checked against the hashes of scripts that reside in the reference utxos that a user specifies in the pre-init config, again, from the release notes. 

If a user wants to compile, export, and deploy their own version, then yes, they switch to the developer workflow, which is as follows:
- Modify scripts, export them, and deploy
- There are a couple of tests that check that the current version of scripts is exported - and exported properly - in case someone missed exporting them.

So the short answer is just run trose test. Shall we make them a separate target? Let me do that, it won't hurt.


> I think we should add a flag to the export utilities that strips out debugging messages from the scripts for production.
We don't use many right now in the happy-path, but it is extremely useful to have more verbose output when doing more exhaustive testing of onchain code. This could be good to add while the code is fresh in your mind.

Makes total sense, there is a flag in the compiler, let me open a ticket - https://linear.app/gummiworm-labs/issue/GUM-97/add-compilation-options-to-script-export

PR #420, opened 2026-04-09
Feat: deploy tx builder

euonymos:
- [ ] Use native scripts explicitly to obtain burn addresses
- [ ] Add the CLI target to run the deplotment

### DIFF
--- a/src/main/scala/hydrozoa/rulebased/ledger/l1/script/plutus/DeploymentTx.scala
+++ b/src/main/scala/hydrozoa/rulebased/ledger/l1/script/plutus/DeploymentTx.scala
@@ -22,20 +22,20 @@ import scalus.cardano.txbuilder.TransactionBuilderStep.{Send, Spend}
   * @param refScriptUtxo
   *   utxo id where the script is located
   */
-final case class DeployTx(
+final case class DeploymentTx(
     scriptDeployed: ScriptHash,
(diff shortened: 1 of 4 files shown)


src/main/scala/hydrozoa/rulebased/ledger/l1/script/plutus/DeploymentTx.scala
```diff
@@ -0,0 +1,109 @@
...
+            val burnAddress =
+                if config.network.isMainnet
+                then
+                    Address.fromBech32(
+                      "addr1wxa7ec20249sqg87yu2aqkqp735qa02q6yd93u28gzul93ghspjnt"
+                    )
+                else
+                    Address.fromBech32(
+                      "addr_test1wza7ec20249sqg87yu2aqkqp735qa02q6yd93u28gzul93gvc4wuw"
+                    )
```
dragospe, line 73: Primarily for audit purposes, I think we should provide either:

- A URL to where you got these addresses, or
- Our own `AlwaysFail` script

dragospe (commented):
Everything seems plausibly correct, LGTM

src/main/scala/hydrozoa/rulebased/ledger/l1/script/plutus/DeploymentTx.scala
euonymos: Ok, let me actually add our own native script that allows spending before slot 0.

PR #426, opened 2026-04-16
Resurrect integration tests

euonymos:


integration/src/test/scala/hydrozoa/integration/stage1/Suite.scala
```diff
@@ -1,691 +1,721 @@
...
+
+    override def genInitialState(env: Env): Gen[State] = {
+
+        logger.trace(s"env start time: ${env.startTime}")
+        import env.testPeers
+        val testPeerToUtxos = env.genesisUtxo(testPeers)
+
+        // Build custom HeadConfig generator with the environment's start time
+        val generateHeadStartTime: GenWithTestPeers[BlockCreationEndTime] =
+            ReaderT(tp => Gen.const(BlockCreationEndTime(env.startTime.quantize(tp.slotConfig))))
```
dragospe, line 305: `ReaderT(tp => Gen.const(....)` can just be replaced with `pure`

src/test/scala/hydrozoa/Generators.md
```diff
@@ -77,5 +73,175 @@ def arbitraryFoo = generateFoo()
...
+- ✅ **Plug and play** - Just provide the part you want to customize at the top level
+- ✅ Simple and direct - TestPeers is explicitly passed
+- ✅ Clear what context is needed - you see `testPeers` at the call site
+- ✅ Concise at the call site - just pass a lambda for context-dependent sub-generators
+
+**Cons:**
+- ❌ Parameter explosion - Top-level generator has many parameters for all customizable parts
+- ❌ Boilerplate in generator implementation - Every generator needs to manually extract `cardanoNetwork` from `testPeers`
+- ❌ Inconsistent - Some context via curried params, some via function types
+- ❌ Not composable - Can't easily chain generators without manually threading TestPeers
```
dragospe, line 160: I don't think this makes a good case for removing `ReaderT`. We know that `TestPeers` contains a `CardanoNetwork`, which contains a `SlotConfig`, and this should be passed implicitly because it never changes. 

Regarding the rest, I do agree that my change pushes the complexity to the call site.

PR #432, opened 2026-04-27
feat: scalus-emulator-based cardano backend

sae3023:
This is a bit of a POC of how the Scalus Emulator can fit into existing hydrozoa code.

Couple of points:

1) The existing Emulator traits seem to cover the basic network interaction well enough: querying utxos, querying chain info and submitting transactions. Therefore, it's possible to omit the `CardanoBackend` trait.
2) The existing application-specific functionality (for now, only the `def lastContinuingTxs`) can be moved to another, higher level trait.

As such, the integration seems pretty seamless.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -37,7 +37,10 @@ dockerCommands := dockerCommands.value.flatMap {
     case other                => List(other)
 }
 
-val scalusVersion = "0.15.1"
+ThisBuild / resolvers +=
+    "Sonatype OSS New Snapshots" at "https://central.sonatype.com/repository/maven-snapshots/"
+
+val scalusVersion = "0.16.0+262-3314a0df-SNAPSHOT"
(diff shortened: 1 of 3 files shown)


euonymos:
@sae3023 I like the idea of splitting up the API into blockfrost-ish standard API and the Hydrozoa-specific part. On the other hand, if we are using Scalus'emulator, doesn't it make sense to eventually just use `BlockchainProviderTF` in Hydrozoa's code?

I see that `BlockchainProvider` pins `BlockchainProviderTF[Future]` - can we somehow start using IO there? Then we can use it directly.

Two things that we need to move forward:
* Updating time/slot when submitting - `IO.realTimeInstant` should be used, maybe a separate handler like setSlot/setTime would be useful as well (though we don't use it)
* Logging - minimally the logger (see Logging.scala), ideally - our new contra-tracer #436

PR #434, opened 2026-05-01
Feat: remove settling sleeps from innerIO, time-precise outerIO

euonymos:


dragospe:
I think it makes sense at a high-level, quite similar to how I understood it working before with some minor tweaks.

I'd need to study it more closely to give a full review. If you're confident in it, I'll defer and move back to my other tasks; if not, I'm happy to give more detailed feedback.

PR #441, opened 2026-05-08
Peter/tracer integration

dragospe:
Wrapped this up to move on to RBR


euonymos:
I like arrows, though I don't see immediately whether it fully solves our performance concerns and particularly your test case with "this should not appear" - I didn't test it, but I guess we still evaluate the argument here:

```
def traceWith(a: A)(using Monad[M]): M[Unit] = this.runTracerA.run(a)
//                ^ strict          
```

But the contramap part should be lazy now because of the squelching.

PR #442, opened 2026-05-08
Peter/rbr model test

dragospe:


src/main/scala/hydrozoa/multisig/ledger/l1/tx/FallbackTx.scala
```diff
@@ -191,6 +191,7 @@ private object FallbackTxOps {
                             mkVoteUtxo(
                               VD.default(treasuryUtxoSpent.datum.commit).toData,
                               config.collectiveContingency.defaultVoteDeposit
+                                  + config.collectiveContingency.minAdaForTreasury
```
euonymos, line 194: Sorry, why minAdaForTreasury is used fot a voteUtxo? Should we add a separate thing for that?

src/main/scala/hydrozoa/rulebased/ledger/l1/tx/EvacuationTx.scala
```diff
@@ -206,16 +214,25 @@ private object EvacuationTxOps {
                     .left
                     .map(BuilderError(_))
 
-                finalized <- context
+                inputValue = inputTreasuryUtxo.treasuryOutput.value
+                outputValue = residualValue
+                evacValue = evacuationOutputs.head.value
```
euonymos: Are those used anywhere?

src/main/scala/hydrozoa/rulebased/ledger/l1/utxo/RuleBasedTreasuryUtxo.scala
```diff
@@ -63,6 +63,37 @@ object RuleBasedTreasuryUtxo {
...
+      *
+      *   - 0 matching UTxOs → `Left(LookupError.Missing)`
+      *   - 1 matching UTxO → parsed via [[parse]]; parse failures become
+      *     `Left(LookupError.ParseFailed)`
+      *   - >1 matching UTxOs → `Left(LookupError.MultipleFound)`
+      *
+      * All `LookupError` variants extend `Throwable` so callers can raise unrecoverable cases with
+      * `IO.raiseError` without wrapping.
+      */
+    def parseOrMissing(
```
euonymos, line 76: Just `parse`?

src/main/scala/hydrozoa/rulebased/DisputeActor.scala
```diff
@@ -317,7 +390,10 @@ object DisputeActor {
     type Handle = ActorRef[IO, Requests.Request]
 
     object Requests {
-        type Request = HandleDisputeRes
+
+        case object PreStart
+
+        type Request = HandleDisputeRes | PreStart.type
```
euonymos: Do we need it? Who sends that?

src/main/scala/hydrozoa/rulebased/DisputeActor.scala
```diff
@@ -394,31 +468,50 @@ object DisputeActor {
...
+        val TreasuryIsUnresolved = "Treasury is Unresolved"
+
     // Parsing. Its in EitherT over IO because we have some recoverable failures (Lefts) and some unrecoverable failures
     // thrown as exceptions.
     // - If we get more than one treasury token or a parsing failure, thats an exception
     // - if we get zero treasury tokens, it may mean that the deinit has succeeded. But we keep trying, in case
     //   of rollbacks.
-    // TODO: Factor out. Its shared between this and the li
-    // obtained from the parameters to this class
     def parseRBTreasury(
```
euonymos, line 430: Can we remove IO here? And combine it with .parseOrMissing?

src/main/scala/hydrozoa/rulebased/DisputeActor.scala
```diff
@@ -223,10 +266,19 @@ final case class DisputeActor(
...
+            _ <- Tracer.routeLocal(s"DisputeActor.${config.ownHeadPeerNum}")
+        } yield ()
+
+    override def receive: Receive[IO, Requests.Request] = {
+        case _: Requests.PreStart.type => preStartLocal
+        case _: Requests.HandleDisputeRes =>
+            handleDisputeRes
     }
 
     /** Queries the cardano backend for all utxos at the dispute resolution address, and then parses
```
euonymos, line 277: I think it's not about parsing utxo but rather about finding a utxo to vote. The description mentions cardano backend but it's not used in fact.

src/main/scala/hydrozoa/rulebased/DisputeActor.scala
```diff
@@ -223,10 +266,19 @@ final case class DisputeActor(
     }
 
     override def preStart: IO[Unit] =
-        context.setReceiveTimeout(config.evacuationBotPollingPeriod, ())
+        context.self ! Requests.PreStart
+
+    private def preStartLocal: IO[Unit] =
+        for {
+            _ <- context.setReceiveTimeout(config.evacuationBotPollingPeriod, ())
```
euonymos: Should be `HandleDisputeRes` not `()`.

src/main/scala/hydrozoa/rulebased/DisputeActor.scala
```diff
@@ -112,10 +143,17 @@ final case class DisputeActor(
...
+                        )
+                        res <- IO.raiseError(NoSuitableCollateralUtxosFound)
+                    } yield res)
             }
-
+            _ <- EitherT.liftF(Tracer.debug("Found collateral utxo"))
         } yield CollateralUtxo(collateralUtxoTuple._1, collateralOutput)
+    }
 
     def handleDisputeRes: IO[Either[DisputeActor.Error.RecoverableErrors, Unit]] = {
```
euonymos, line 155: I am not fully convinced the term "dispute utxos" make sense. How are they different from "vote utxos"?

src/main/scala/hydrozoa/rulebased/DisputeActor.scala
```diff
@@ -163,6 +203,8 @@ final case class DisputeActor(
                     val keySorted = otherUtxos.sortBy(_.voteOutput.key)
                     val continuing = keySorted.head
                     for {
+                        _ <- EitherT.liftF(Tracer.info("Tallying..."))
+
                         removed <- keySorted.tail.find(voteUtxo =>
                             voteUtxo.voteOutput.key == continuing.voteOutput.link
                         ) match {
```
euonymos, line 205: Arguably this is part of parsing. Does it make sense to introduce a sum type to represent possible outcomes more systematically?

src/main/scala/hydrozoa/rulebased/EvacuationActor.scala
```diff
@@ -2,43 +2,68 @@ package hydrozoa.rulebased
...
+  * @param evacuationMapAtFallback
+  * @param fallbackTxHash
+  * @param tracerLocal
+  * @param config
+  */
 case class EvacuationActor(
-    toEvacuate: EvacuationMap,
+    thisNodeEvacuates: EvacuationMap,
     cardanoBackend: CardanoBackend[IO],
     evacuationMapAtFallback: EvacuationMap,
```
euonymos, line 48: Ultimately this should match the resolved version, not the one when the fallback happened.

src/main/scala/hydrozoa/rulebased/EvacuationActor.scala
```diff
@@ -2,43 +2,68 @@ package hydrozoa.rulebased
...
+
     override def preStart: IO[Unit] =
-        context.setReceiveTimeout(config.evacuationBotPollingPeriod, ())
+        context.self ! Requests.PreStart
 
-    override def receive: Receive[IO, Requests.Request] = { case _: Requests.Evacuate =>
-        handleEvacuation
+    private def preStartLocal: IO[Unit] =
+        for {
+            _ <- context.setReceiveTimeout(config.evacuationBotPollingPeriod, ())
```
euonymos: Should be `Evacuate` not `()`

euonymos (approved):

src/main/scala/hydrozoa/rulebased/EvacuationActor.scala
```diff
@@ -128,9 +164,20 @@ case class EvacuationActor(
...
+                )
+
+                // Evacuations _we_ still have to attempt
+                toEvacuate: EvacuationMap =
+                    thisNodeEvacuates.removedAll(previouslyEvacuated)
 
+                _ <- EitherT.liftF(
+                  if toEvacuate.isEmpty
+                  then Tracer.info(LogMessages.NoMoreEvacuations)
+                  else Tracer.info(s"${toEvacuate.size} payout obligations left")
```
euonymos, line 180: Let's stop calling it payout obligations please.

src/main/scala/hydrozoa/rulebased/ledger/l1/tx/EvacuationTx.scala
```diff
@@ -135,18 +137,24 @@ private object EvacuationTxOps {
                   left = EvacuationTx.Build.Error.NoEvacuatees
                 )
 
+                // The subset we will actually evacuate in this transaction (capped at 63).
```
euonymos: I see only take(64) in line 144, should it be 64 in the comment?

src/main/scala/hydrozoa/rulebased/ledger/l1/utxo/VoteUtxo.scala
```diff
@@ -12,17 +12,72 @@ import hydrozoa.rulebased.ledger.l1.script.plutus.DisputeResolutionValidator.Dis
...
+            }
+            voteDatum <- Try(fromData[VoteDatum](inline.data)) match {
+                case Success(d) => Right(d)
+                case Failure(e) => Left(ParseError.DatumDeserializationError(utxo, e))
+            }
+            tokenCount <- utxo.output.value.assets.assets
+                .get(config.headMultisigScript.policyId)
+                .flatMap(_.get(config.headTokenNames.voteTokenName))
+                .toRight(ParseError.MissingVoteToken(utxo))
+            voteTokens = PositiveInt.unsafeApply(tokenCount.toInt)
```
euonymos: Is it ok to use unsafe* here? Maybe it's ok, but it should be logged properly if this happens.

src/test/scala/hydrozoa/rulebased/ledger/l1/tx/EvacuationTxTest.scala
```diff
@@ -210,28 +194,36 @@ object EvacuationTxTest extends Properties("EvacuationTx Test") {
...
+            },
             "Evacuated outputs should match builder's evacuatees"
             // Note this holds because we're not
             // testing large numbers of evacuations
+
           )
           _ <- assertWith(evacuationTx.tx != null, "Transaction should not be null")
 
           // Verify residual treasury value is correct
           totalEvacuations =
```
euonymos, line 218: This seems to be unused.

euonymos:
I can't remember where I saw it exactly, but anyways: liquidation -> evacuation everywhere in names and comments.

src/main/scala/hydrozoa/multisig/ledger/l1/tx/FallbackTx.scala
dragospe: From here:

https://discord.com/channels/1465451842451275968/1473725778503143506/1506891651048669274

>The default vote utxo provides the minADA sufficient for an empty-liabilities treasury, which gets spent by deinit.

src/main/scala/hydrozoa/rulebased/ledger/l1/tx/EvacuationTx.scala
dragospe: Nope, they're left over from debugging logs that have since been removed. I wonder why the linter didn't flag them as unused?

src/main/scala/hydrozoa/rulebased/ledger/l1/utxo/RuleBasedTreasuryUtxo.scala
dragospe: Well, this signature is distinct from `parse` on line 56, including a different error type. We could overload `parse`, but I thought it'd be a bit strange.

Which do you prefer?

src/main/scala/hydrozoa/rulebased/DisputeActor.scala
dragospe: `Prestart` is (or rather, will be) sent by the Rulebased Regime Manager on initialization, which does context.setReceiveTimeout -- then the timeout itself sends ().

I'll update to use the type alias in `preStartLocal` instead of just `()`.

src/main/scala/hydrozoa/rulebased/DisputeActor.scala
dragospe: No, because the line on 495 (`case Left(e) => throw e`) should actually be `IO.raiseError(e)`, I believe.

src/main/scala/hydrozoa/rulebased/DisputeActor.scala
dragospe: Good catch, this is a stale comment.

dragospe: Although, this function _is_ specifically assuming that the presence of the vote token has already been checked.

src/main/scala/hydrozoa/rulebased/DisputeActor.scala
dragospe: I don't mind changing the name, but what I had in mind was:

- "dispute utxo" :: something at the dispute resolution script. May or may not parse into a valid utxo
- "vote utxo" :: something that carries the correct vote token and correctly parses into a `VoteUtxo`

Basically just pre vs. post parsing. I'd eventually put these into types, even, where the cardano backend has a smart constructor to give us `DisputeUtxos`, then `VoteUtxos` get parsed from that, etc.

src/main/scala/hydrozoa/rulebased/DisputeActor.scala
dragospe: I think this comment might have been intended to be on more lines than just this. Are you suggesting that we do a sum type for the `Vote`, `Tally`, `Resolve` cases in the `_ <- disputeUtxos match {(...)}` block?

If so, I agree -- that will make things more clear.

src/main/scala/hydrozoa/rulebased/ledger/l1/tx/EvacuationTx.scala
dragospe: It should, yes. 

Actually, this should be a parameter somewhere and not a magic number, as we discussed prior to my leave

src/main/scala/hydrozoa/rulebased/EvacuationActor.scala
dragospe: Sure. I recall having this discussion, but I don't remember what we settled on.


Also, the type in the `EvacuationMap` is called `Payout.Obligation` -- should we change that as well?

src/main/scala/hydrozoa/rulebased/ledger/l1/utxo/VoteUtxo.scala
dragospe: No, I don't think this is okay any more.

src/main/scala/hydrozoa/rulebased/ledger/l1/tx/EvacuationTx.scala
dragospe: I put it in EvacuationTx.Assumptions for now, but I think the entire KZG thing should be wrapped into something higher level soon.

PR #452, opened 2026-05-25
The fully integrated branch for slow consensus

euonymos:
This integrates several branches:
- fast consensus split
- slow consensus impl
- rate limiter
- moving KZG into SEC-effect
- stage1/stage4 upgrades


src/main/scala/hydrozoa/multisig/consensus/transport/Codecs.scala
euonymos, line 1: I am leaving this without a thorough review for now:
* https://linear.app/gummiworm-labs/issue/GUM-131
* https://linear.app/gummiworm-labs/issue/GUM-132

PR #458, opened 2026-06-04
Peter/rbr manager

dragospe:
> NOTE: This targets `feature/persistence-recovery` because it requires access to the RocksDB setup, but the target branch still needs to backmerge `main`.

  ## Summary

  Resurrects the RuleBasedRegimeManager and brings the rule-based regime end-to-end:  boot from persistence, dispatch the dispute flow on a typed `DisputeAction`, and add  the on-chain `Abstain` branch as a peer's escape hatch when no SEC is available.

  ### What's in here

  - **`RuleBasedRegimeManager`** boots from persistence: walks `Markers.hardConfirmed`  → `StoreKey.HardConfirmation` → `LaneKey.Stack[lastBlockNum]` →    `StoreKey.EvacuationMap[lastBlockNum]` (per §5.2 / §6 of   `design/persistence-and-crash-recovery.md`) to recover `(action, evacMap,  fallbackTxHash)`. `Initial` stack yields `Abstain` + `config.initialEvacuationMap`;     `Regular` yields `Vote(sec, sigs)` when the latest partition carries a SEC,
    `Abstain` otherwise.
  - **`DisputeAction` sum** — `Vote(sec, signatures) | Abstain` — replaces the prior
    `(sec, signatures)` constructor pair on `DisputeActor`. Vote builds `VoteTx`,
    Abstain builds `AbstainTx`.
  - **On-chain Abstain** — new `DisputeRedeemer.Abstain` branch in
    `DisputeResolutionValidator`, off-chain `AbstainTx` builder, `VoteOutput.abstain`
    transition, regenerated `plutus.json` blueprint.
  - **`TallyTx`** skips `ValidityStartSlot` when both tally inputs are decided (Voted
    or Abstain) — matches the on-chain `isAwaiting` guard.
  - **`EvacuationActor`** treats `NoEvacuatees` as recoverable so the actor stays
    alive for rollbacks after its last evacuation lands.
  - **`DisputeActor` cleanup** — `buildAndSubmit` collapses the build/log/sign/submit
    pattern across the four arms; `DisputeUtxos.Resolve` now carries `VoteUtxo[Voted]`
    with a parse-time check (`NonVotedUtxoAtResolve` covers spec violations).
  - **Latent fixes:** `VoteTx.voteUtxoProduced` index 0→1 (off by one in the output
    list); `DeploymentTxTest` funding 50→100 ADA for the larger script.
  - **Integration test** — `EvacuationPropertyTest` covers both Vote and Abstain
    happy paths (parameterized scenario), spawns `DisputeActor` + `EvacuationActor`
    directly per peer (RBRM-driven integration deferred until a stage4-style
    multisig harness is available), and waits 30s after the terminal log fires so
    any post-completion crash surfaces.
- **EvacuationActor** replaces `State.evacuationMapAtFallback` with a runtime query to find the datum of the initial resolved treasury in order to compute the active `EvacuationMap`. `RuleBasedRegimeManager` pre-loads all known `candidateEvacuationMaps`.

  ### Out of scope / follow-ups

  - Integration test through `RuleBasedRegimeManager` (the persistence-backed boot
    path) is deferred until a stage4-style harness can drive a head through
    hard-confirmation + fallback.
- Prune `candidateEvacuationMaps` to only those maps which are _after_ the fallback (voting will reject anything with a lower version)


dragospe:
Current implementation uses the _last_ fallback, but its not guaranteed that this is the fallback that was submitted. We need to get that transaction and proceed in the RBRM accordingly.

dragospe:
For transitioning from MR to RBR:
- Panic shuts down everything except cardano liason
- Starts RBRM

PR #459, opened 2026-06-05
Tracing Migration

dragospe:
Migrates all actors (infrastructure code still uses the old `loggerIo`)  into `ContraTracer`, with `Slf4jTracer` being a type alias/companion object for `ContraTracer[IO, LogEvent]` and routing keys set by the consumer.

All actors now trace typed events (see: `JointLedgerEvent.scala`) which are captured and formatted by various callers (see `JointLedgerEventFormat.scala`). `IOLocal[Tracer]` has gone away; the actors are now ignorant of SLF4J. 

This also subsumes `ProtocolTracer`, which is now just another formatter (and emitted through the SLF4J logger).

The primary motivation of this work was to remove `BlockBriefObserver` and `StackObserver` from stage4 in order to facilitate using an actual multisig regime manager (rather than the manually-wired actors) as a pre-requisite to testing the multisig regime -> fallback -> rule-based regime flow.


euonymos:
`docs/integration-stages.md` might need to reflect the changes in stage4's observability tools

PR #465, opened 2026-06-11
integer request id

rrruko:
This changes the default encoding of RequestId to use the integer representation. I have not taken care to ensure that this change is limited to the interface with the L2; please let me know if that is a problem.


euonymos:
Superseded by branch `feature/gum-131` (two commits): one unifies RequestId on the single canonical object shape `{headPeerNumber, requestNumber}` everywhere (closes GUM-131), and one adds a *scoped* `RequestId.i64` codec opted into only at the L2-ledger sites (`L2LedgerCommand`/`RemoteL2LedgerCodecs` + `L2StoreCodecs`) — so SugarRush gets the u64 form without leaking i64 into the consensus wire, persistence journals, or the HTTP API. Your `ApplyTransaction` decoder fix (read `userVk`) is ported as-is. Suggest closing this in favor of that branch.

PR #466, opened 2026-06-11
populate missing NodeConfig fields from .env

rrruko:


euonymos:
This should come from the config, not the env. From the env, you should only know you peerId - the rest should be taken from the config.

PR #476, opened 2026-06-18
stage4 + Main on MRM: file-based config; PeerTransport trait + in-process variant

dragospe:
## Summary

Thirteen commits, in four arcs:

### Arc A — file-based config + Main rewiring

| SHA | What |
|---|---|
| `6adef0ea` | `feat(config): add GenerateSampleConfig CLI tool with round-trip test` |
| `938d7252` | `refactor(mrm): rename PeerWsTransport -> PeerTransport; inject as Resource[ctx => PeerTransport]` |
| `503b9cbf` | `refactor(main): load NodeConfig from files; drop env-vars + Bootstrap.mkNodeConfig` |

`Main` becomes a `CommandIOApp` taking `<head-config.json> <peer-private.json>`. Everything the node needs at runtime (Blockfrost key, wallet keys, bind host/port, Sugar Rush WS URI, HTTP admin creds) moves into `NodePrivateConfig`; the process reads no env-vars. `Bootstrap.mkNodeConfig` (~250 LOC UTxO-fetching + init-tx-building) is removed from the runtime path — configs are produced by `GenerateSampleConfig` (or a future tool). `Migrate` + `TokenRecovery` siblings rewired as `CommandIOApp`s taking the same args.

Fixed a pre-existing bug in `NodeConfig.fromJson`: it was reading `cardanoNetwork`/`headPeers` from `.headConfigBootstrap.X` paths that the encoder never writes.

`Main.runNode` gains optional `dataDir` / `httpExtraTracer` / `backendOverride` params for testability. `MainSmokeTest` exercises end-to-end startup to `HydrozoaHttpEvent.ServerStarted` with a mocked backend + ephemeral ports.

### Arc B — stage4 head peers onto MRM

| SHA | What |
|---|---|
| `e0427475` | `refactor(transport): extract PeerTransport trait; WS impl is WsPeerTransport` |
| `e9c2fcb5` | `feat(transport): add InProcessPeerTransport for in-process multi-peer tests` |
| `47eaeeef` | `test(stage4): disable coil-mixed properties; pending MRM port of coil flow` |
| `45e388fd` | `test(stage4): add MRM-based sutResource path for Direct + no-coil scenarios` |
| `32103392` | `test(stage4): port WS-mode tests onto MRM; delete PeerStack + buildPeerStack` |

`PeerTransport` is now an interface; `WsPeerTransport` is the real WS impl (renamed) and `InProcessPeerTransport` routes sends directly to local liaison handles via a shared `Registry`. Stage4's `sutResource` calls `HeadMultisigRegimeManager.resource` per peer instead of hand-wiring 8 actors via `buildPeerStack` + `postStack` + manual `Connections`. The new `Observers` object captures `SCA.StackHardConfirmed` and `JL.BriefProduced` events through the HMRM event sum (`|+|` combined with the slf4j sink).

**Net:** stage4 `Suite.scala` shrank by ~1000 lines; deleted dead code: `buildPeerStack`, `postStack`, `transportSetup`, `buildCoilWirings`, `setupWebSocketNetwork`, old `acquireSut`, and the `CoilWiring` / `WsNetwork` / `PeerStack` case classes.

Coil-mixed tests were temporarily disabled (`Prop.passed` with TODO) in `47eaeeef`; restored in Arc C below.

### Arc C — coil wiring + MRM refactors

| SHA | What |
|---|---|
| `a76b0baf` | `feat(coil): restore coil wiring on MRM; extract base + tracer carrier; ephemeral test ports` |

Coil flow restored end-to-end on the MRM-based wiring path:

- **New abstract transports**: `HubTransport` / `CoilTransport` traits (in `HubWsTransport.scala` / `CoilPeerWsTransport.scala`) + `InProcessHubCoilTransport` (Hub + Coil sub-classes sharing one registry) for Direct-mode tests. `RemoteCoilProxy` / `RemoteHubProxy` now depend on the traits, not the WS impls.
- **HMRM** gains optional `hubCoilTransport: Option[Resource[IO, ctx => HubTransport]]`. When this peer has hubbed coils, the MRM internally registers each `PeerLiaisonHubToCoil` on the transport and spawns one `RemoteCoilProxy` per hubbed coil, populating `Connections.remoteCoilLiaisons`. Mirrors the existing head-mesh wiring pattern.
- **CMRM** gains `coilTransport: Resource[IO, ctx => CoilTransport]` parameter + `.resource(...)` factory (replaces dead `.apply`). Spawns `RemoteHubProxy` and populates `Connections.remoteHubLiaison`.
- **Stage4 Suite**: `headNetworkFor` builds mesh + optional hub-coil transports on one shared `NodeWsServer`; new `coilUplinkFor` + `buildCoilMrm`; main for-comprehension allocates the `InProcessHubCoilTransport.Registry`, builds CoilMRMs, merges coil ticks/stacks into the Sut. `Observers.captureCoilStackHardConfirmed` collects per-coil hard-confirmed stacks.
- **Main.scala**: dispatches head vs coil from `nodeConfig.ownPeerId`. New `buildHeadNode` mounts mesh + (when hub) hub routes on one shared NodeWsServer; new `buildCoilNode` dials the hub's `/hub`. Coil node skips the HTTP server (no user-facing requests).
- **Re-enabled** 3 disabled properties in Runner: `Two-heads-one-coil works` (Direct + extended) and `Two-heads-one-coil works WS`.

Refactors on top of the coil restoration:

- **`MultisigRegimeManagerBase`** (new file): shared `supervisorStrategy`, `preStart`, `receive` dispatch, `connectionsDeferred`, `watchChildren` helper, and a `spawnCoreActors` method that allocates the six actors present on every multisig peer (BW/CL/FCA/JL/SC/SCA). HMRM and CMRM extend the base and only implement `preStartLocal` for their role-specific pieces. Removes ~80 LOC of duplicated scaffolding across the two MRMs.
- **`MrmTracers` carrier** (alongside the `HeadMultisigRegimeManagerEvent` ADT): one case-class field per event variant + a `fromRoot(rootTracer)` factory that does all the `contramap` routing in one place. Each subclass holds a `lazy val tracers: MrmTracers` instead of individual `xTracer` lazy vals. Variant names + carrier field names spelled out (`BWL` → `BlockWeaverLimiter`, `bw` → `blockWeaver`, etc.).
- **OS ephemeral ports for tests**: `WsPeerTransport.create` takes `remoteIds: List[HeadPeerId]` (no URIs); `startDialers` takes the URI map. `CoilPeerWsTransport.startDialer` takes the hub URI at dial-start time. Stage4 binds NodeWsServer with port 0, captures `server.address.getPort` post-bind, then computes the URI map and starts dialers in a second pass. `TransportMode.WebSocket` drops its `basePort` field. Fixes a `BindException` collision between two sequential WS-mode tests in the same JVM. **Production paths unchanged** — Main still passes `nodeConfig.hydrozoaPort` for the bind, so deployers' firewall rules apply as before.

**Net for Arc C**: 16 files, +1134 / −490. Two new files (`MultisigRegimeManagerBase.scala`, `InProcessHubCoilTransport.scala`); HMRM + CMRM net shrink ~80 LOC each from base extraction; stage4 Suite + Main grow to accommodate the coil branches.

### Arc D — settling: signal-based wait for CL submission

| SHA | What |
|---|---|
| `8831a08a` | `refactor: rename Tx trait -> EnrichedTx` |
| `6ebcd951` | `refactor(cardano-liaison): actionTxs returns Seq[EnrichedTx]` |
| `8b8b35a3` | `test(stage4): effectsLandedSignal — signal-based wait for CL submission` |
| `27b853d8` | `todo(cardano-liaison): narrow Action ADT element types` |

CI on the post-Arc-C tree failed `Two-heads-one-coil works WS` (run `28026757851`) with *"Falsified after 0 passed tests"* — `EffectsLanded.propEffectsLanded` ran while every backbone tx was still unknown to the backend.

**Root cause.** `RateLimits.default.hardStackMinPeriod = 3.minutes` (production cadence). In Direct/TestControl mode the throttle is bridged by virtual time; in WS mode it is real wall-clock and the test's actor budget is only ~30s. `Stack.HardConfirmed` sits in the `Limiter` between SCA and StackComposer for 3 real minutes, so CL never receives `PushResults` inside the test window and submits nothing. `slowCoverageSignal` fires anyway because it is anchored on `SlowConsensusActorEvent.StackHardConfirmed` — emitted by SCA *before* the limiter — so the suite proceeds straight into `analyzeBlockBriefs` with the mock backend still empty.

**Fix.** A second harness signal, anchored on the *enactment* event (`CardanoLiaisonEvent.TxSubmitting`) so the suite blocks until the same condition `propEffectsLanded` accepts is satisfied. `slowCoverageSignal` stays as the consensus-reach barrier; the new `effectsLandedSignal` fires when each backbone `BlockExpectation` is satisfied by either its happy txs or its competing fallback. Plus a tractable test-only `hardStackMinPeriod = 2.seconds` (down from 3 min) so the throttle still batches but fits the WS wall-clock budget — `softBlockMinPeriod` left at the production default so block-batching cadence is unchanged.

Foundation work for the signal (Arc D's first two commits, requested independently while debugging):

- **`Tx[Self <: Tx[Self]]` renamed `EnrichedTx`** (`8831a08a`, 45 files): the project's wrapper around the raw `scalus.cardano.ledger.Transaction` was named ambiguously. `\bTx\b` sweep + scalafmt — `TxTiming` / `TxSignature` / `signTx` / `txLens` are unaffected (no word-boundary match). Prose-comment collateral (`Tx-body`, `Tally Tx Test`, etc., where the contraction was for "transaction" not the trait) reverted to lowercase `tx` or kept as the contraction.
- **`CardanoLiaison.actionTxs` returns `Seq[EnrichedTx[?]]`** (`6ebcd951`): the `Action` ADT now carries `EnrichedTx[?]` end-to-end instead of stripping to raw `Transaction` at every construction site. Raw `Transaction` only appears at the `cardanoBackend.submitTx` boundary. `CardanoLiaisonEvent.TxSubmitting` enriched to carry `TransactionHash` directly (was `String`) so the harness signal can match without round-tripping through `.toString`. A TODO commit (`27b853d8`) flags the obvious follow-up — element types are still `EnrichedTx[?]` existentials where construction sites already hold concrete subtypes (`FallbackTx` / `RolloutTx` / `HappyPathEffect`); narrowing makes the `.tx` strip a compile error.

**Net for Arc D**: 7 files (signal wiring) + 45 files (rename, mostly mechanical). Signal-wiring net is +157 / −10.

## Test plan

- [x] `just build-werror` clean (main + integration)
- [x] `just fmt` + `just lint` clean
- [x] `MainSmokeTest` passes (reaches `HydrozoaHttpEvent.ServerStarted`)
- [x] `GenerateSampleConfigTest` round-trip (100 generated configs) still passes
- [x] `CoilConfigGenTest` still passes
- [x] Core unit tests pass (274 tests, 0 failures)
- [x] Stage4 `Two-peers head works` Direct (100/100 in 26s) on post-Arc-D tree — `effectsLandedSignal` table confirms 4/4 backbone txs land Happy per scenario with the new 2s `hardStackMinPeriod`
- [ ] `Two-heads-one-coil works WS` in CI — the failure this PR's Arc D fixes; awaiting green CI on the pushed branch

## Notes for reviewers

- **History was force-pushed**: branch was rebased onto current `main` and grew from 1 commit (the old PR `218bb95c`) to 13 commits. Re-checkout if you had the old tip.
- **Migrate / TokenRecovery rewrites are untested end-to-end.** They compile and follow the same load-from-config pattern as `Main`, but there's no smoke test exercising them.
- **`InProcessPeerTransport.send` silently drops messages to unregistered peers** during the per-peer setup race (peer 0's children start sending before peer N's transport exists). Recovered by `PeerLiaisonHeadToHead`'s `GetMsgBatch` protocol on top — same recovery path that masks WS dialer transient failures.
- **Stage4 settling flakes pre-date this PR.** Two head-only properties (`Twenty-peers head works`, the diagnostic `Two-peers 1 command`) failed with `stack coverage: 1 observed block(s) never in any hard-confirmed stack` in an earlier prepush; the fix landed independently on `main` as `0b816e72` and is now part of this branch via the rebase. Coil-mixed tests inherit the same settling story — re-enabling them is a wiring restoration, not a guarantee they'll never flake. Arc D narrows one specific class of WS-mode flake (timing race against the StackComposer throttle), not the broader settling story.
- **Test-only `RateLimits` override in stage4** (`hardStackMinPeriod = 2s`) is via the existing `NodeOperationMultisigConfigGen` seam — production defaults in `RateLimits.default` are untouched.
- **Heavy merge conflict expected with PR #480** (`ilia/multi-peer-main-bootstrap`, off main). That PR rewrites Bootstrap + Main for production multi-peer dispatch using concrete `WsWiring` carriers; this branch uses abstract trait factories. Same intent on HMRM/CMRM, different shape. PR #480 also flips `Connections.requestSequencer` from `Option` to mandatory (with a `NoopActor` on coil); this branch keeps `Option`.
- **Arc D adds a `Tx` → `EnrichedTx` rename touching 45 files.** Mostly mechanical (verified by the 92% file-similarity git rename detection on `Tx.scala` → `EnrichedTx.scala`); see Arc D for the regex strategy. Will conflict with any in-flight branch that touches the `multisig.ledger.l1.tx` package — easiest to rebase forward by re-running the same sweep.


integration/src/test/scala/hydrozoa/integration/stage4/Runner.scala
```diff
@@ -172,7 +165,7 @@ object Stage4Properties extends YetAnotherProperties("Integration Stage 4"):
     val _ = property("Two-peers head works WS") = Stage4Suite(
       label = "stage4-ws-two-peers",
       nPeers = 2,
-      transportMode = TransportMode.WebSocket(),
+      transportMode = TransportMode.WebSocket,
```
dragospe, line 168: Switched to ephermal port numbers to prevent races

dragospe:
## Pre-existing: `Twenty-peers head works (extended)` poll-cadence divergence

Found during local prepush (extended set, not in `integration-fast`). Failing seed: `Vxa2fZuhM3pJneAgM88eQXfJbLIOwUMK3oUGktw951M=`

**Error:**
```
Reference block brief didn't match actual block brief; consensus is broken.
actual:   depositsAbsorbed: List((14,0),(8,0),(19,0),(0,1),(15,3)),       depositsRefunded: List((7,0),(9,0),(4,4))
expected: depositsAbsorbed: List((14,0),(8,0),(19,0),(0,1),(15,3),(4,4)), depositsRefunded: List((7,0),(9,0))
```

**Root cause:** Deposit `(4,4)` was classified as `NotInPollResults` (→ refunded) on the follower but `Eligible` (→ absorbed) on the leader. Both agreed the deposit was mature. The divergence is `isExistent`: the leader's `CardanoLiaison` had already seen the deposit UTxO in its poll results; the follower's hadn't yet. This is the "poll-cadence divergence" scenario documented at `JointLedger.scala:810–816`.

**Not introduced by this branch** — the extended 20-peer stress test creates more concurrent deposits and more polling races. The regular `Twenty-peers head works` (10 peers, `integration-fast`) passes 10/10 on this branch after the `slowCoverageTarget` fix from `0b816e72`.

src/main/scala/hydrozoa/config/node/NodePrivateConfig.scala
```diff
@@ -15,6 +15,11 @@ final case class NodePrivateConfig(
     override val hydrozoaHost: String,
     override val hydrozoaPort: String,
     override val blockfrostApiKey: String,
+    override val sugarRushUri: String,
+    override val adminUsername: String,
+    override val adminPassword: String,
+    override val httpHost: String,
+    override val httpPort: String,
```
dragospe: These should not all be strings. use proper URI types where applicable

PR #477, opened 2026-06-18
fix(rulebased): pin treasury output address in Evacuate

rssh:
## Summary

`RuleBasedTreasuryValidator`'s **`Evacuate`** branch pins the continuing treasury output's beacon token, datum fields and total value — but never its **address**. The `Resolve` branch checks the address; `Evacuate` does not.

Because `Evacuate` is **permissionless** (outputs are meant to be pinned by the KZG evacuation commitment), any third party can redirect a *resolved* treasury to an address they control.

### Exploit (empty-subset evacuation)

An `Evacuate` with `evacuationKeys = []` and `proof = <current accumulator>`:

- passes the membership check — with an empty subset, `e(acc, g₂) == e(proof, g₂)` reduces to `acc == proof`;
- has `evacuatedValue = 0`, so value-preservation `treasuryInput.value === treasuryOutput.value + evacuatedValue` **forces the entire treasury** (beacon + all backing funds) into the continuing output;
- …which the attacker simply addresses to their own wallet.

Every other check (beacon present, datum fields preserved, value preserved) still passes, so the whole treasury is stolen. Partial evacuations are affected too — the residual treasury that sits between batches can be redirected the same way.

## Fix

One `require`, mirroring the `Resolve` branch:

```scala
require(treasuryOutput.address === treasuryInput.address, EvacuateTreasuryWrongAddress)
```

## Test

`RuleBasedTreasuryScriptTest` gains a regression test that drives the **real** on-chain `Evacuate` logic (the BLS pairing runs on the JVM via `blst`). It submits two byte-for-byte identical transactions, varying only the treasury-output address:

- → treasury **script** address: accepted (proves the tx is otherwise valid);
- → **attacker** address: **rejected** by the new check.

Before the fix, the second case was accepted — the live vulnerability.

## Verification

- `sbtn "testOnly *RuleBasedTreasuryScriptTest"` → green (5 passed, 1 pre-existing `ignore`d).
- `scalafmtCheck` / `Test/scalafmtCheck` → clean; pre-commit `lint fmt check` passed.

> Heads-up unrelated to this PR: `just build-werror` currently reports 57 pre-existing `-Werror` "unused value of type Assertion" errors in `multisig/{consensus,persistence/recovery}` test files (from #467) under JDK 25. None are in the files this PR touches.

### DIFF
--- a/src/main/scala/hydrozoa/rulebased/ledger/l1/script/plutus/RuleBasedTreasuryScript.scala
+++ b/src/main/scala/hydrozoa/rulebased/ledger/l1/script/plutus/RuleBasedTreasuryScript.scala
@@ -112,6 +112,8 @@ object RuleBasedTreasuryValidator extends Validator {
         "version in treasuryInput and treasuryOutput must match"
     private inline val EvacuateSetupG2ShouldBePreserved =
         "setupG2 in treasuryInput and treasuryOutput must match"
+    private inline val EvacuateTreasuryWrongAddress =
+        "Treasury output must remain at the treasury script address"
 
     // Deinit redeemer
     private inline val DeinitRequiresResolvedTreasury =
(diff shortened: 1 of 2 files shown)


euonymos:
Seuperseded by #492.

PR #478, opened 2026-06-18
fix(rulebased): reject zero-evacuatee Evacuate (no-progress guard, L-01)

rssh:
## Summary

Follow-up to #477, **stacked on it** (see note below). Closes the **L-01** finding from the security review: a permissionless *zero-evacuatee* `Evacuate` is a valid no-op that just re-creates the resolved treasury.

### The DoS

The resolved treasury is a single global UTxO that every evacuation must spend. An empty-subset `Evacuate` (`evacuationKeys = []`, `proof = current accumulator`) passes all checks and re-creates the treasury unchanged — for the price of a tx fee and **no signature**. An attacker can race such no-ops to invalidate in-flight honest evacuations block-by-block, stalling everyone's exit during the wind-down. No funds are lost (availability, not integrity), hence Low severity — but it's free griefing.

## Fix

```scala
require(!evacuationKeys.isEmpty, EvacuateMustMakeProgress)
```

The existing `evacuationOutputs.size == evacuationKeys.size` check ties outputs to keys, so every `Evacuate` must now remove ≥1 committed utxo. Because evacuation outputs are hash-pinned to the rightful owners' committed addresses/values, the only way to spend the treasury is to actually pay one out — so a would-be griefer's spend *advances* the protocol instead of stalling it. Re-evacuating an already-removed utxo fails the membership check, so there's no replay churn. Consistent with the off-chain builder, which never emits an empty evacuation (`NoEvacuatees`).

Inherent (symmetric) contention from the single-treasury design remains — a separate design matter (sharding), out of scope here.

## Tests

- **Reworked the C-01 regression test** onto a *real* single-utxo evacuation (active set `{evacuee}` → empty residual, `proof` = G1 generator), so the address check is exercised independently of the new guard. (The old C-01 test was built on the empty no-op, which this guard now rejects.)
- **New L-01 test:** the empty (no-progress) evacuation — otherwise valid — is rejected.

Both drive the real on-chain `Evacuate` (BLS pairing via `blst`); `testOnly *RuleBasedTreasuryScriptTest` → 6 passed, 1 pre-existing ignored.

## ⚠️ Stacked on #477

This branch is based on #477's branch, so until #477 merges this PR's diff shows **both** commits. The L-01-specific change is commit `2d6d997c`. Please review/merge **after** #477 — GitHub then collapses this to the L-01 diff automatically.


euonymos:
Superseded by #493.

PR #503, opened 2026-06-30
test(rulebased): co-spend two ballot boxes in one Vote tx (E6)

rssh:
## What

Adds **E6** to `DisputeVoteAttackTest` — a regression test for the co-spend guard
(`VoteOnlyOneVoteUtxoIsSpent`) in the rule-based dispute Vote validator.

E6 builds a single Vote tx that co-spends **two** ballot boxes with **different source
txids** and asserts the builder's Plutus evaluator rejects it with the guard's message. A
`buildTwoBoxVoteTx` helper mirrors `VoteTx.Build`'s step list with two `votingSpend` + `send`
pairs (both boxes ratchet to the same `sec`, so they share one Vote redeemer; both named peers
sign).

## Why

The "only one vote utxo" guard used to filter co-spent inputs by `outRef.id` (the source-tx hash)
only. That missed two already-ratcheted boxes whose source txs diverge from the FallbackTx. The
on-chain audit (`8b63c6fe`) replaced it with a token-identity filter —
`i.outRef !== ownRef && i.resolved.value.containsCurrencySymbol(headMp)` — which catches a second
ballot box (or the treasury) regardless of source txid. E6 locks that behaviour in: if the guard
ever regresses to a txid filter, E6 goes red.

## Test

`sbt "testOnly *DisputeVoteAttackTest"` → green (E1–E6 pass).


src/test/scala/hydrozoa/rulebased/ledger/l1/script/plutus/DisputeVoteAttackTest.scala
```diff
@@ -109,6 +126,62 @@ object DisputeVoteAttackTest extends Properties("Dispute Vote Attack") {
...
+              )
+            )
+            finalized <- context
+                .addRequiredSigners(boxA.votingSigners ++ boxB.votingSigners)
+                .finalizeContext(
+                  diffHandler = contextualscalus.Change.changeOutputDiffHandler(0),
+                  validators = nonSigningValidators
+                )
+        } yield finalized.transaction
+    }
```
dragospe, line 184: Perhaps we should modify the tx builders to expose their transaction `context` at the end, so that  tests can modify the context directly rather than duplicating the structure?

src/test/scala/hydrozoa/rulebased/ledger/l1/script/plutus/DisputeVoteAttackTest.scala
```diff
@@ -278,6 +351,58 @@ object DisputeVoteAttackTest extends Properties("Dispute Vote Attack") {
...
+        e6BoxA <- failLeft(
+          BallotBox
+              .parse(e6BoxAUtxo)(using config)
+              .map(_.asInstanceOf[BallotBox[VoteStatus.AwaitingVote]])
+        )
+        e6BoxB <- failLeft(
+          BallotBox
+              .parse(e6BoxBUtxo)(using config)
+              .map(_.asInstanceOf[BallotBox[VoteStatus.AwaitingVote]])
+        )
```
dragospe, line 388: Not blocking, but generally I would prefer something like

```scala
        e6BoxA <-
            for 
                e6TxidA <-  pick(Arbitrary.arbitrary[TransactionHash])
                e6BoxAUtxo <- mkBallotBoxUtxo(
                   5,
                   6,
                   VoteStatus.AwaitingVote(ownWallet.exportVerificationKey.pubKeyHash),
                   TransactionInput(e6TxidA, 0)
                )
              res <- failLeft(
                  BallotBox
                     .parse(e6BoxAUtxo)(using config)
                     .map(_.asInstanceOf[BallotBox[VoteStatus.AwaitingVote]])
                   )
            yield res
```

It makes it more clear that `e6TxidA` and `e6BoxAUtxo` are only intermediate values and are irrelevant outside of the bind scope.

Even better, a helper function `def mkBox(key : Int, link : Int) : Gen[BallotBox]` would make it clear that they're generated in the same way.

dragospe (approved):
Two non-blocking comments, mergable if CI passes

PR #513, opened 2026-07-06
feat(server): L2 utxo + recent-transactions query endpoints

euonymos:
## What

Two read-only L2 query endpoints on the user-facing HTTP server (Fund14 proj68, MS1):

- **`GET /api/l2/utxos/{address}`** — current L2 utxos an address controls, CIP-0116-structured JSON. `400` on a malformed address, `[]` on an unknown-but-well-formed one.
- **`GET /api/l2/transactions?count=N`** — recent applied L2 transactions, newest first. `count` defaults to 50; a non-integer `count` is a `400`.

## Design

- A narrow **`L2LedgerReader[F]`** trait (`utxosByAddress`, `recentTransactions`) that `L2Ledger[F]` extends. `EutxoL2Ledger` implements it (a live `Ref` filter for utxos; a tail-scan of its own command log for recent transactions — **no new RocksDB storage**). `RemoteL2Ledger` **stubs both** (returns empty), following its existing `currentCommandNumber`/`restoreTo` stub pattern.
- The HTTP route holds only the reader, not the writable ledger (least-knowledge).
- Utxos serialize via the repo's existing **CIP-0116** codecs plus one new structured `TransactionOutput` encoder (`{address, value, datum}`; datum tagged `{inline:…}` / `{hash:…}` / `null`). Transaction summaries are `{requestId, blockNumber, kind}` with `kind ∈ transaction | depositRegistered | depositAbsorbed | depositRefunded`.

## EUTXO-only

The queries are answered by the **EUTXO reference ledger**, which holds its state locally. A production node currently wires **`RemoteL2Ledger`** (the SugarRush WebSocket), which owns its state behind the black box — so on such a node both endpoints return an empty array (a valid, non-error response). This matches the milestone's "formally works only with the EUTXO ledger."

## Demo / tests

`multisig/server/L2QueryEndpointsTest` (8 tests) boots an in-memory `EutxoL2Ledger`, seeds it, builds the real routes, and drives both endpoints over HTTP — asserting the responses and printing the JSON, so the flow can be shown in a screen recording. Covers the populated case, empty-address `[]`, malformed-address `400`, `count` limit / default / `0`, malformed `count` `400`, and the CIP-0116 shape.

## Verification

- Main + integration compile clean; `just fmt` / `just lint` clean.
- 67 ledger / recovery / server / smoke tests pass (0 failures), plus the 8 new endpoint tests.
- An 18-agent adversarial review found only convention/nit issues (no correctness bugs); 6 were applied (404→400 count validation, count-by-summary windowing, tagged datum, `toBech32.getOrElse(toHex)`, expanded test coverage).

## Docs

`docs/l2-query-endpoints.md`, indexed in `docs/README.md` and `CLAUDE.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -90,6 +90,10 @@ Working design and reference docs live in `docs/`:
 - [`rate-limiter.md`](docs/rate-limiter.md) — generic throttling actor that slows the consensus
   cycles without changing consensus logic.
 
+**API**
+- [`l2-query-endpoints.md`](docs/l2-query-endpoints.md) — the user-facing server's read-only L2
+  queries: `GET /api/l2/utxos/{address}` and `GET /api/l2/transactions` (EUTXO-only).
+
 **Testing**
 - [`integration-stages.md`](docs/integration-stages.md) — the stage1/stage4 integration test
   levels: what each exercises and where to add a test.
--- a/build.sbt
+++ b/build.sbt
@@ -40,6 +40,7 @@ dockerCommands := dockerCommands.value.flatMap {
 val scalusVersion = "0.18.1"
 val bloxbeanVersion = "0.7.1"
 val http4sVersion = "0.23.32"
+val tapirVersion = "1.13.25"
 
 // Cardano on-chain validators and shared on-chain types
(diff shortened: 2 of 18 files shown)


euonymos:
Included in #524

PR #518, opened 2026-07-07
test(build): fork test JVM with parallel suites + JDK-25 native access + concurrency cap

rssh:
## What

Configure the `core` test task to fork a JVM, keep suites parallel, silence JDK-25 warnings, and cap concurrency for CI:

```scala
Test / fork := true
Test / testForkedParallel := true
Test / javaOptions ++= Seq("--enable-native-access=ALL-UNNAMED", "--sun-misc-unsafe-memory-access=allow")
Global / concurrentRestrictions := Seq(Tags.limitAll(6), Tags.limit(Tags.ForkedTestGroup, 1))
```

## Why each piece

- **`Test / fork := true`** — isolates native state (RocksDB JNI), the cats-effect `IORuntime`, and daemon threads across runs; also makes re-running the same `testOnly` in a warm sbt session actually re-run (without fork it silently reports "0 tests" until a recompile intervenes).
- **`Test / testForkedParallel := true`** — forking otherwise **serializes** the suites. Measured full `sbt test`: no-fork **26:03** → fork-serial **69:06** → fork+parallel **21:52**. This flag restores parallel suite execution.
- **`Test / javaOptions` (native-access flags)** — forked test JVMs don't inherit the sbt launcher's JDK-25 flags, so they emit `sun.misc.Unsafe` (Scala `LazyVals`) and `System::load` (blst-java JNI) restricted-method warnings. These grant the same access and silence them.
- **`Global / concurrentRestrictions` → `Tags.limitAll(6)`** — the suites are wall-clock/sleep-bound (real `System` timestamps in the cats-actors), so overlapping their waits wants **more** concurrent suites than cores. sbt's default caps concurrency at core count (**2** on GitHub runners), which leaves those waits un-overlapped. Pinning to 6 lets 6 sleep-bound suites overlap on a 2-core runner. On a 12-core dev machine it's ~free: **22:16** with the cap vs **21:52** unbounded.

## Validation

Full `sbt test` inside `nix develop`: **292 / 292 green, 0 failures, 0 JDK-25 warnings** (timing 22:16). The 2-core CI benefit is best read from this PR's own CI test duration.

https://claude.ai/code/session_01A18n2gt9sRkb1vMuzKEMtX


dragospe:
_Most likely_ the CI failures are due to race conditions that have been fixed on `main`. If CI passes with this last backmerge, this can get merged in -- otherwise, I'll need to investigate more closely.

PR #519, opened 2026-07-07
rbr: review followups

dragospe:
- Timing tuning for dispute resolution tests
- Less unsafe IO usage
- Adds an in-memory HTTP submission client instead of using the request sequencer actor handle directly, exercising the signed JSON request API
- Adds back in actual evacuation (not just resolution) in EvacuationPropertyTest.scala and updates classifier
- Adds CIP30 signing to the wallet module for the HTTP endpoint + associated tests
- Adds setupG2 as non-empty to the production path for the fallback treasusry (previously the evac test was doing this synthetically)
  - adds the minAda to the unresolved treasury output instead of the public ballot box (!). If the L2 value was low, the inclusion of the G2 datum to unresolved treasury could bump the treasury above the minAda threshold.
  - Why did we put the minAda for the treasury in the public ballot box?
  - This goes away in the next ticket I'll be working on, probably
- Spends the collateral to pay the fee for evacuation txs rather than requiring another fee utxo. Simplifies things.
- Improves tracing and pattern matching in the RBA
- Gates tally on voting deadline for better tracing
- Adds "evacuation" sentinel datums to init parameters gen for UTXO classifier purposes

### DIFF
--- a/src/test/scala/hydrozoa/config/head/initialization/InitializationParametersGen.scala
+++ b/src/test/scala/hydrozoa/config/head/initialization/InitializationParametersGen.scala
@@ -270,16 +270,25 @@ object InitializationParametersGenTopDown {
                         for {
                             next <- generateCappedValue(cardanoNetwork)(capValue = rest)
                             address <- Gen.oneOf(peerAddresses)
-                            acc_ = acc :+ Babbage(
+                            // The datum bumps the output's min-ADA above what generateCappedValue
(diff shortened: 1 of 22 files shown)


dragospe:
NB: requires changes from unpushed work

src/test/scala/hydrozoa/config/head/initialization/InitializationParametersGen.scala
dragospe, line 1: Is this still necessary?

PR #522, opened 2026-07-08
Route Stage4 harness submissions through in-memory HTTP JSON endpoints

dragospe:
## Summary

Integrates the production HTTP JSON endpoints into the multipeer harness so
Stage4 submissions exercise the real request path — JSON encode + CIP-30 COSE
sign + http4s round-trip + COSE verify + `RequestSequencer` — instead of
sending a `UserRequest` message directly to the peer's actor. No sockets:
each peer's `HydrozoaRoutes` is wrapped in `http4s.Client.fromHttpApp`, so it
composes with TestControl-mode direct runs unchanged.

Changes in order:

- `JsonCodecs`: add the previously-commented `Decoder[RequestAccepted]`,
  mirroring the encoder's `i64` shape.
- Introduce `SubmissionClient` trait (`submit(UserRequest): IO[RequestId]`)
  with two impls: `.direct(handle)` (actor `?:`) and `.http(client, uri,
  wallet)` (signs the header CIP-30, routes deposits vs L2 tx to the
  matching endpoint, decodes `RequestAccepted`).
- `WalletModule`: add abstract `signCoseCip30`; each impl (BloxBean, Scalus)
  keeps its private key encapsulated and internally derives the enterprise
  address of `blake2b_224(pubKey)` for bloxbean's `CIP30DataSigner.verify`
  to walk back to on the server side.
- `MultiPeerHeadHarness.Peer[H]` gains a `submissionClient` field; a new
  `Http.mkPeerSubmissionClient` builds per-peer `HydrozoaRoutes`, wraps
  them via `Client.fromHttpApp`, and hands back a `SubmissionClient.http`
  bound to the peer's own wallet.
- Stage4 `Suite` / `Sut`: drop the now-empty `Stage4PeerHandle` wrapper;
  `Stage4SutStatic.peers: Map[HeadPeerNumber, SubmissionClient]` reads
  each client from `harness.peers(n).submissionClient`.

## Test plan

- [x] `sbtn compile` clean on both core and `integration / Test / compile`
- [x] `integration/testOnly Stage4Properties -z "Two-peers"` — the extended
      variant (500 commands) passes 100 tests over the http round-trip;
      earlier failure `AddressRuntimeException: Unknown network type` was
      the trigger for the enterprise-address derivation fix
- [x] Full `Stage4Properties` suite (non-extended, extended, WS variants)
- [x] `EvacuationPropertyTest` + `VoteVersionMismatchTest` still compile
      (they carry their own `H` type, so the added `Peer[H].submissionClient`
      field is transparent)

### DIFF
--- a/integration/src/test/scala/hydrozoa/integration/harness/MultiPeerHeadHarness.scala
+++ b/integration/src/test/scala/hydrozoa/integration/harness/MultiPeerHeadHarness.scala
@@ -27,11 +27,14 @@ import hydrozoa.multisig.consensus.transport.*
 import hydrozoa.multisig.ledger.eutxol2.EutxoL2Ledger
 import hydrozoa.multisig.persistence.rocksdb.RocksDbBackendStore
 import hydrozoa.multisig.persistence.{BackendStore, Cf, InMemoryBackendStore, Persistence, PersistenceEvent, PersistenceEventFormat}
+import hydrozoa.multisig.server.{HydrozoaHttpEvent, HydrozoaHttpEventFormat, HydrozoaRoutes, HydrozoaServer, SubmissionClient}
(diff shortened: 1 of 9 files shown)


dragospe:
Folded into #519 (combined PR). Superseded.

PR #525, opened 2026-07-09
feat(rbr): full evacuation property test; EvacuationTx pays fee from collateral

dragospe:
## Summary

- `FallbackTx.setupG2`: use `TrustedSetup.takeSrsG2` instead of `SList.empty` so the RBR evacuation flow can actually run end-to-end.
- `InitializationParametersGen`: stamp L2 outputs with an `"evacuation"` inline-datum sentinel so `RBRClassifier` can bucket the resulting L1 evacuation outputs by content (the sentinel is committed to the KZG at obligation-creation time, so the on-chain membership check reproduces the same hash and accepts the L1 output).
- `EvacuationTx`: pay the fee from the collateral utxo (`add + spend + collateralOutput.send`, diff at index 0), matching the pattern used by `VoteTx` / `ResolutionTx` / `DeinitTx` et al. Drops the separate `feeUtxos` parameter and the wallet-input Spend loop.
- `RuleBasedActor`: `getFeeAndCollateral` → `getEvacuationCollateral`, returns just `CollateralUtxo`.
- `EvacuationPropertyTest`: asserts the full-drain terminal state.
  - `EvacuationOutputPlaceId` is the first `PayoutsLeft(n)` observed on any peer's RBA (the KZG-committed evacuation count at read-of-treasury time).
  - `AmbientPlaceId = nHeadPeers * 2` — each peer maintains two parallel wallet-ADA streams (init-change + fallback-equity); every RBR-side script tx that pays fee from collateral picks one of these, returns it fee-subtracted at the same address, so no utxo is destroyed.
- Test-file cleanups for the `feeUtxos` removal (`EvacuationTxTest`, `EvacuationAttackTest`, `EvacuationDrainTest`, `L2OutputEvacuabilityTest`).

## Test plan

- [x] `just fmt-check`
- [x] `just lint-check`
- [x] `just build-werror`
- [x] `sbtn "integration/testOnly *EvacuationPropertyTest*"` — passes with `EvacuationOutputPlaceId == PayoutsLeft(n)` and `AmbientPlaceId == 4` for the 2-peer scenario.
- [ ] Existing evacuation unit tests (`EvacuationTxTest`, `EvacuationAttackTest`, `EvacuationDrainTest`, `L2OutputEvacuabilityTest`) still pass after the `feeUtxos` parameter removal.

### DIFF
--- a/integration/src/test/scala/hydrozoa/integration/rbr/property/EvacuationPropertyTest.scala
+++ b/integration/src/test/scala/hydrozoa/integration/rbr/property/EvacuationPropertyTest.scala
@@ -13,6 +13,7 @@ import hydrozoa.integration.harness.MultiPeerHeadHarness
 import hydrozoa.integration.harness.MultiPeerHeadHarness.Transport.Mode as TransportMode
 import hydrozoa.lib.cardano.scalus.QuantizedTime.QuantizedInstant
 import hydrozoa.lib.logging.{ContraTracer, Slf4jTracer}
+import scala.annotation.unused
 import hydrozoa.multisig.backend.cardano.{CardanoBackend as L1Backend, FirewalledCardanoBackend, FirewalledCardanoBackendEvent, yaciTestSauceGenesis}
 import hydrozoa.multisig.consensus.peer.{HeadPeerNumber, PeerId}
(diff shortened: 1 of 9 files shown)


dragospe:
Folded into #519 (combined PR). Superseded.

PR #528, opened 2026-07-10
Evict setupG2 and immutable head-identity fields from the rule-based treasury datum

dragospe:
## What

The rule-based treasury datum previously inlined a 65-element compressed-G2 vector (`setupG2`, ~6.2 KB) and the immutable head-identity fields (`headMp, disputeId, headPeers, headPeersN, coilPeers, coilQuorum`), bloating every treasury transition and forcing preservation checks in the validator.

- **setupG2 → deployed-once setup ladder.** Seven inline-datum utxos at the burn address, outputs 0–6 of a single deployment tx. Self-contained prefixes of sizes {2,3,5,9,17,33,65}: rung *i* holds the first 2^i+1 points and covers exactly 2^i evacuations. An `EvacuationTx` references the smallest covering rung and points the validator at it via a new `setupRefInputIdx` redeemer field. `maxEvacuationsPerTx` stays 64.
- **Immutable fields → rule-based regime utxo.** The FallbackTx *moves* the regime witness beacon to a new output at the **head multisig (native script) address** carrying a `RuleBasedRegimeDatum` — no new plutus script: any spend already requires the unanimous multisig, which is the deinit condition. It is distinguished from the datum-less multisig regime utxo at the same address by its datum, and validators locate it as a reference input by the beacon under `headMp`.
- **Ladder authentication.** The regime datum records the ladder's anchor (rung 0's full outRef), written from `ScriptReferenceUtxos.setupLadderInputs` by the all-peers-signed FallbackTx. Evacuate requires its setup reference input to satisfy `outRef.id === anchor.id && idx < 7`, so a degenerate/foreign setup datum is rejected (covered by a live negative test).
- **HRWT rename.** The multisig-regime witness token is now the "Hydrozoa regime witness token": CIP-67 tag 4679 (HMRW) → 4798 (HRWT), prefix `012be4e0`.

The treasury datum shrinks to `Unresolved(headMp, deadlineVoting, versionMajor)` / `Resolved(headMp, evacuationActive, version)` (154 / 204 CBOR bytes measured, was ~6.9 KB); the onchain/offchain datum split and `TreasuryStateConversions` are gone. `headMp` stays in the datum as the induction anchor for the regime-utxo lookup — the treasury holds arbitrary L2 assets, so the validator cannot derive it from its own value.

## Notable follow-ons

- Treasury and dispute script hashes changed → existing deployed reference scripts are invalidated; `DeployReferenceScripts` now chains a third tx deploying the ladder and writes the extended `script-refs.json`.
- `FallbackContingency.Collective` gains `minAdaForRegime`.
- "HRWT must be burned when the regime utxo is spent" is convention (DeinitTx does it), not script-enforced — any spend requires the same unanimous multisig that deinit requires.

## Testing

- Full unit suite including reworked onchain-script tests (regime/ladder reference inputs), builder tests, size tests, DeinitTx spending the regime utxo, and a negative test that a byte-identical rung datum at a foreign tx id is rejected by the anchor check.
- Integration petri model gains `RegimeRefPlaceId` / `SetupLadderRefPlaceId` places with classifier rules and terminal-cardinality assertions.

## Folded in from #533

This branch also carries the former #533 (`worktree-rbr-commitment-selection`), now closed:

- **Voting-deadline fall-through (`RuleBasedActor`).** Once the voting deadline has elapsed, own `VoteTx`/`AbstainTx` (head) and `RatchetVoteTx` (coil) can no longer land, so both dispute branches now gate on `hasDeadlineElapsed` and dispatch to the residual tally/resolve path instead, tracing `VotingDeadlineElapsed`.
- **Evacuation walk-back fix.** `loadEvacuationInputs` no longer errors with `MissingState("walked back to 1 …")` on a minor-only stack 1; it decrements to stack 0's `Initial`, which terminates the walk with the initial fallback. This was the crash behind the #528 CI timeout.
- **`CommitmentSelectionPropertyTest`.** New integration property (Test 1: firewall votes → default major-1 evacuates; Test 2 stubbed). Its expected cardinalities include the `RegimeRefPlaceId` / `SetupLadderRefPlaceId` places this PR introduces.

The merge threads this PR's `regimeUtxo` reference through the deadline-gated dispute branches.


cardano-onchain/src/main/scala/hydrozoa/rulebased/ledger/l1/state/RegimeState.scala
```diff
@@ -0,0 +1,32 @@
...
+      * (the datum itself is trusted because the regime utxo is produced by the all-peers-signed
+      * FallbackTx).
+      */
+    case class RuleBasedRegimeDatum(
+        disputeId: TokenName,
+        headPeers: List[VerificationKey],
+        headPeersN: BigInt,
+        coilPeers: List[VerificationKey],
+        coilQuorum: BigInt,
+        setupLadder: TxOutRef
```
dragospe:
```suggestion
        setupG2Ladder: TxOutRef
```

cardano-onchain/src/main/scala/hydrozoa/rulebased/ledger/l1/state/TreasuryState.scala
dragospe, line 1: - Add comment that we need to keep the headMp for the resolution.
- See if we can remove it from `Resolved`
- Move it to a `def` on the enum and override in the cases if not

src/main/scala/hydrozoa/app/DeployScriptsAndG2Setup.scala
dragospe, line 1: Change name to "DeployScriptsAndG2Setup"

dragospe: We deploy the setup ladder exactly once (the G2 setup doesnt change); the scripts may get updated on new release. Make sure we're not just chaining all three together every time.

src/main/scala/hydrozoa/config/head/multisig/fallback/FallbackContingency.scala
```diff
@@ -171,5 +177,9 @@ object FallbackContingency {
         // Max serialized size of a rule-based treasury utxo when there are no L2 liabilities
         // (see TreasuryOutputSizeTest)
         val maxNoLiabilitiesTreasuryUtxoBytes: PositiveInt = PositiveInt.unsafeApply(6960)
+
+        // Max serialized size of the rule-based regime utxo (HRWT beacon + head-identity datum,
+        // dominated by the peer vkey lists)
+        val maxRegimeUtxoBytes: PositiveInt = PositiveInt.unsafeApply(1024)
```
dragospe, line 183: Validate this number, I think its a hallucination.

src/main/scala/hydrozoa/config/HydrozoaBlueprint.scala
dragospe, line 1: This can be undone.

src/main/scala/hydrozoa/rulebased/ledger/l1/script/plutus/DeploymentTx.scala
```diff
@@ -51,48 +54,56 @@ private object DeploymentTxOps {
...
       *   key-locked. The change is sent back to the address of the first utxo.
-      * @param scriptToDeploy
-      *   the script we are deploying
+      * @param payloads
+      *   the payloads we are deploying, one utxo each
       */
     final case class Build(
         utxosToSpend: NonEmptyList[Utxo],
-        scriptToDeploy: ScriptRef
+        payloads: NonEmptyList[DeployedPayload]
```
dragospe, line 74: This should probably be `scriptPayload : Option[ScriptRef]` and `dataPayload: Option[ScriptRef]`

cardano-onchain/src/main/scala/hydrozoa/rulebased/ledger/l1/state/TreasuryState.scala
dragospe: Ensure that all redeemers that spend the treasury utxo keep the headMp immutable

dragospe: Implement transition functions to update the treasury datums in only valid ways (keeping the correct fields immutable)

cardano-onchain/src/main/scala/hydrozoa/rulebased/ledger/l1/state/RegimeState.scala
dragospe: Done in 1db8a070 — renamed the datum field to `setupG2Ladder` (RegimeState) and its use sites (RuleBasedTreasuryScript, RuleBasedRegimeUtxo).

cardano-onchain/src/main/scala/hydrozoa/rulebased/ledger/l1/state/TreasuryState.scala
dragospe: Done in 1db8a070. It can't be dropped from `Resolved` (both Evacuate and Deinit read it), so headMp is now an abstract `def headMp` on the enum — a uniform accessor implemented by each case's field — with a doc comment on why it's preserved.

src/main/scala/hydrozoa/app/DeployScriptsAndG2Setup.scala
dragospe: Done in b08d5122 — renamed the object/file and CLI (`deploy-scripts-and-g2-setup`); updated the Bootstrap doc refs and the justfile recipe.

dragospe: Done in b08d5122 — added `--ladder-refs <existing script-refs.json>`: when given, the deployed ladder is reused and only the two validator scripts are redeployed. The ladder tx is built only in the bootstrap (no `--ladder-refs`) case.

src/main/scala/hydrozoa/config/head/multisig/fallback/FallbackContingency.scala
dragospe: Validated in 648fbd54 — the new RegimeOutputSizeTest measures the regime output at the maximum head (20 peers) at 877 bytes, so 1024 holds as a ceiling with headroom (~one compressed vkey per peer). Kept 1024 and grounded the comment in the measurement.

src/main/scala/hydrozoa/config/HydrozoaBlueprint.scala
dragospe: Reverted in 6efe906e — HydrozoaBlueprint back to main.

src/main/scala/hydrozoa/rulebased/ledger/l1/script/plutus/DeploymentTx.scala
dragospe: Done in b08d5122 — `DeployedPayload` is now a case class with `scriptPayload: Option[ScriptRef]` and `dataPayload: Option[Data]`, so a single deployed utxo can carry both a ref script and an inline datum (with `.script`/`.data` convenience constructors).

cardano-onchain/src/main/scala/hydrozoa/rulebased/ledger/l1/state/TreasuryState.scala
dragospe: Done in 1db8a070 — Resolve and Evacuate now check the output datum's headMp against the transition result; Deinit is terminal (burns all head tokens, no continuing treasury output, so nothing to preserve).

dragospe: Done in 1db8a070 — added `Unresolved.resolve(...)` and `Resolved.evacuate(...)` in TreasuryState. The validator builds the expected next datum from these and checks each output field against it (single source of truth for what advances vs. stays immutable); the off-chain builders (ResolutionTx, EvacuationTx) construct via the same methods.

PR #533, opened 2026-07-13
Worktree rbr commitment selection

dragospe:


dragospe:
Folded into #528 — that branch now carries this one's deadline fall-through, evacuation walk-back fix, and CommitmentSelectionPropertyTest.

PR #535, opened 2026-07-14
test(integration): bump takeoffOffset 10s -> 60s so head init survives slow bring-up

dragospe:
## Problem

Integration-fast runs on main (and PR branches, e.g. #518's CI run) are failing with scenario timeouts in the real-clock multi-peer harness suites. The logs show every peer's `CardanoLiaison` warning:

```
init tx validity window elapsed (currentTime >= initializationTxEndTime);
head can no longer be initialized on the happy path
```

`VoteVersionMismatchTest`, `EvacuationPropertyTest`, and `CommitmentSelectionPropertyTest` all bring up their heads with `takeoffOffset = 10.seconds`. When suites run concurrently (or on a slow 2-core CI runner), harness bring-up takes longer than 10s, the init tx validity window elapses before the leader can submit it, the head never initializes, and the scenario times out.

Reproduced locally on unmodified `origin/main` (99d85fb6) with the CI command `sbt "integration/testOnly * -- -s 10 -f ^(?!.*\(extended\))"`: same init-window warnings across the 2-, 3-, and 5-peer suites, ending in `RBR Evacuation Property.ws: ... Exception: TimeoutException: 5 minutes`.

## Fix

Bump `takeoffOffset` to 60 seconds in all three suites, giving bring-up a 60s budget before the init window closes. The per-step scenario timeouts (90s / 5min) are unaffected; the block/dispute schedule is anchored to takeoff time, so steps just start later.

Local verification run with the bump is in progress; will report results on this PR.

### DIFF
--- a/integration/src/test/scala/hydrozoa/integration/fallbackhandoff/VoteVersionMismatchTest.scala
+++ b/integration/src/test/scala/hydrozoa/integration/fallbackhandoff/VoteVersionMismatchTest.scala
@@ -87,7 +87,7 @@ object VoteVersionMismatchTest extends Properties("Vote Version Mismatch"):
           transportMode = transportMode,
           testPeers = testPeers,
           testPeerToUtxos = testPeerToUtxos,
-          takeoffOffset = 10.seconds,
+          takeoffOffset = 60.seconds,
           disputeResolutionConfig = fastDisputeResolutionConfig,
           coilPeers = coilPeersConfig,
         ) { (takeoffTime, mnc) =>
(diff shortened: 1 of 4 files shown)


dragospe:
Local verification results (running the CI command `sbt "integration/testOnly * -- -s 10 -f ^(?!.*\(extended\))"`):

**Unmodified origin/main (99d85fb6)** — reproduces the failure: init-window warnings across the harness suites, `RBR Evacuation Property.ws` failed with `TimeoutException: 5 minutes` (seed `FwciGW_TkSlxt1leMpu5wBwcJSulVI9WVwZQ1dW4y4L=`).

**With this branch (takeoffOffset 60s)** — `Vote Version Mismatch` and `RBR Evacuation Property` now pass. `CommitmentSelectionPropertyTest` still failed in the full-parallel local run (13 suites at once on a many-core machine — far more contention than CI's 2-core runner), but passes when run alone (96s). So the remaining local failure is CPU starvation specific to high local parallelism, not expected on CI.

Mechanism note: `initializationTxEndTime = blockCreationEndTime (= takeoff) + minSettlement (2s) + inactivityMargin (10s)` — `takeoffOffset` only buys bring-up runway *before* takeoff; after takeoff the leader has ~12s to submit the init tx. If CI still misses the window, the next knob is `inactivityMarginDuration` in the suites' `widenedTxTiming`, not more takeoff offset.

PR #536, opened 2026-07-16
test(rbr): RBR petri-net model + terminal-agree unit tests (MBT PR1a)

dragospe:
## Summary

- Adds a Scala 3 Petri-net model of the rule-based regime's post-fallback lifecycle (Vote / Abstain / Ratchet / six-flavor Tally / Resolution / Evacuation) plus a canonical-order firing driver (`RBRModel`) that reaches a deterministic terminal marking from post-fallback initial state.
- Adds `RBRModelTest` — eight parameter-scenario terminal-agree cases, three `Params` validation checks, one topology-validity check. 12/12 passing.
- Extends `RBRClassifier` with an `AbstainPlaceId` fn (parse `VoteStatus`).
- Extends the hand-written expected-cardinalities maps in `EvacuationPropertyTest` and `CommitmentSelectionPropertyTest` so the new `AbstainPlaceId` and three synthetic `TallyBuffer*` places don't break equality (all → 0).

This is **PR1a** of the model-based-testing plan for RBR — the model in isolation, no `MultiPeerHeadHarness` wiring yet. PR1b will bring in the sibling \`ModelBasedSuite\` wiring the model's terminal against real-SUT snapshots plus a confluence property test.

## Notable design points

- **Six-flavor Tally split** — \`TallyTx\` is one on-chain family, but the marking-only arc semantics cannot express \"consume any two of {Unvoted, Abstain, Voted}\" atomically. ISO 15909-1 HLPN could collapse these via typed arc inscriptions; note in \`RBRTransitionId\`.
- **Two-step same-status Tally** — PT(2)+TP(1) on the same \`(place, transition)\` pair violates the library's \`SingleArc\` topology constraint: the simulator applies both arcs against the pre-fire snapshot and merges via \`TreeMap\` last-write-wins, giving a wrong net effect. Worked around by adding three synthetic \`TallyBuffer*\` places and splitting each same-status Tally into \`TallyXX\` (PT(2) X → TP(1) buffer) plus \`TallyRestoreX\` (PT(1) buffer → TP(1) X). Restore is ordered before Consume in the canonical firing order so each pair completes atomically.
- **Mixed Tally stays atomic** — \`TallyUA/UV/AV\` use PT(loser) + Read(winner). Different places → no SingleArc conflict.
- **No Deinit** — per tech-lead decision peers submit \`DeinitTx\` manually, so the model terminal matches the RBA's terminal (\`ResolvedTreasury = 1\`, \`EvacuationOutput = nEvacs\`).

## Test plan

- [x] \`sbtn integration/Test/compile\` clean under \`CI=1\` (\`-Werror\`)
- [x] \`sbtn 'integration/testOnly hydrozoa.integration.rbr.model.RBRModelTest'\` — 12/12
- [ ] Existing property suites (\`EvacuationPropertyTest\`, \`CommitmentSelectionPropertyTest\`) still pass — extended the expected map and classifier compile cleanly; not exercised locally because they run the full harness (minutes per scenario)


integration/src/test/scala/hydrozoa/integration/rbr/model/petri/net/transitions/Evacuation.scala
```diff
@@ -3,179 +3,33 @@ package hydrozoa.integration.rbr.model.petri.net.transitions
...
+      RBRArcId.EvacuationReadTreasuryRef ->
+          RBRReadArc(TreasuryRefPlaceId, id, PositiveInt.unsafeApply(1)),
+      RBRArcId.EvacuationReadResolved ->
+          RBRReadArc(ResolvedTreasuryPlaceId, id, PositiveInt.unsafeApply(1)),
+      RBRArcId.EvacuationReadSetupLadder ->
+          RBRReadArc(SetupLadderRefPlaceId, id, PositiveInt.unsafeApply(1)),
+      RBRArcId.EvacuationReadCollateral ->
+          RBRReadArc(CollateralPlaceId, id, PositiveInt.unsafeApply(1)),
+      RBRArcId.EvacuationSpendAmbient ->
+          RBRPTArc(AmbientPlaceId, id, PositiveInt.unsafeApply(2)),
```
dragospe, line 29: Evacuation only spends 1 ambient utxo now, not two. We folded in collateral + fee in a previous commit.

Verify this against the tx builder. And actually, the assertion that the tests pass is a little bit alarming given this -- I feel like this _shouldn't_ pass. Collateral utxos, I believe, also won't have the sentinel datum. Your handling of this in the classifier is, therefore, suspect (but I haven't reviewed that yet so I can't be certain).

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/net/transitions/Abstain.scala
```diff
@@ -0,0 +1,27 @@
...
+    val id: RBRTransitionId = RBRTransitionId.Abstain
+
+    val arcs: List[(RBRArcId, RBRArc)] = {
+        val w1 = PositiveInt.unsafeApply(1)
+        List(
+          RBRArcId.AbstainSpendUnvoted -> RBRPTArc(UnvotedPlaceId, id, w1),
+          RBRArcId.AbstainProduceAbstain -> RBRTPArc(AbstainPlaceId, id, w1),
+        )
+    }
+}
```
dragospe, line 27: abstain SHOULD actually gate on the voting deadline, probably. Lets implement this (in the script, tx builder, RBA, and associated tests) on a branch off main. It doesn't belong in this PR necesarily.

But let's also leave a code comment to reflect that this arc list must update.

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/net/transitions/Ratchet.scala
```diff
@@ -0,0 +1,26 @@
...
+
+    val arcs: List[(RBRArcId, RBRArc)] = {
+        val w1 = PositiveInt.unsafeApply(1)
+        List(
+          RBRArcId.RatchetReadTreasury -> RBRReadArc(UnresolvedTreasuryPlaceId, id, w1),
+          RBRArcId.RatchetSpendAbstain -> RBRPTArc(AbstainPlaceId, id, w1),
+          RBRArcId.RatchetProduceVoted -> RBRTPArc(VotedPlaceId, id, w1),
+        )
+    }
+}
```
dragospe, line 26: We still need a voted -> voted ratchet. That is still a transition that can fire, and if we're comparing transition sequences it is important.

We may want to alter the library a bit to give us something closer to ISO 15909 HLNs. This would probably make vote ratcheting easier to express.

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/net/transitions/Resolution.scala
```diff
@@ -0,0 +1,33 @@
...
+import hydrozoa.integration.rbr.model.petri.net.*
+import hydrozoa.integration.rbr.model.petri.net.RBRPlaceId.*
+import hydrozoa.integration.rbr.model.petri.net.Transitions.RBRTransitionId
+import hydrozoa.lib.number.{NonNegativeInt, PositiveInt}
+
+/** Arcs for [[RBRTransitionId.Resolution]] (ResolutionTx).
+  *
+  * Consumes the unresolved treasury plus one `Voted` ballot box; produces the resolved treasury
+  * plus `nEvacs` tokens on `PayoutObligationsPlace`. The `nEvacs` weight is fixed at net
+  * construction (the winning kzg commitment's evacuation-map size on-chain).
```
dragospe, line 12: > The `nEvacs` weight is fixed at net construction

Perhaps. This may need to be reconsidered. There wouldn't be anything stopping us from updating the net on-the-fly in response to model commands, although this would complicate some analysis techniques.

If we preferred to keep a fixed amount, what we would need to do is generate commands that eventually push L2 state towards an evacuation map with the expected number, and then trigger fallback at that point.

Thoughts?

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/net/transitions/Abstain.scala
dragospe: Confirmed by @GeorgeFlerovsky: only tallying after the deadline.

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/net/transitions/Resolution.scala
```diff
@@ -0,0 +1,33 @@
...
+  *
+  * When `nEvacs == 0` the `ProducePayoutObligations` arc is omitted — a `PositiveInt(0)` is not
+  * constructible — so Resolution still fires but seeds no evacuation work.
```
dragospe, line 15: We should investigate how this is handled in the SUT/ground truth. I seem to recall thinking about this once before, but I don't remember what the outcome was.

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/net/transitions/Tally.scala
```diff
@@ -0,0 +1,130 @@
...
+  * These consume 2 tokens of X and produce 1 token of X (net X -= 1). Encoded via a two-step detour
+  * through a synthetic `TallyBufferXPlaceId`:
+  *   - `TallyXX` — PT(2) X, TP(1) buffer, Read(1) UnresolvedTreasury.
+  *   - `TallyRestoreX` — PT(1) buffer, TP(1) X. No treasury read; it just returns the "winner"
+  *     token to X so the pair completes.
+  *
+  * The detour exists because the library's `SingleArc` topology constraint forbids putting both a
+  * PT and a TP arc on the same (place, transition) pair — the simulator would then apply both
+  * against the same pre-snapshot and merge via last-write-wins on the TreeMap, giving a wrong net
+  * effect.
```
dragospe, line 27: N.B.: The "single arc" constraint was due to semantic ambiguity for non-commutative firings. For instance, if you have a reset arc and a TP(n) arc attached to same place, you either end up with 0 or n tokens depending on which arc "fires" first.

I seem to recall the description in 15909-1 seeming to avoid these issues, and 15909-3 establishing some notion of "orthogonality" describing which types of net extensions were mutually compatible.

This is another point in favor of revising the library closer to 15909.

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/net/Places.scala
```diff
@@ -31,6 +31,12 @@ enum RBRPlaceId:
     case ResolvedTreasuryPlaceId
     case UnvotedPlaceId
     case VotedPlaceId
+    case AbstainPlaceId
+    // Synthetic buffers used only during the two-step same-status Tally (see
+    // transitions/Tally.scala). Marking is 0 at both initial and terminal states.
+    case TallyBufferUnvotedPlaceId
+    case TallyBufferAbstainPlaceId
+    case TallyBufferVotedPlaceId
```
dragospe, line 39: I don't love these. But it is what it is.

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/net/RBRNet.scala
```diff
@@ -0,0 +1,147 @@
...
+      * count.
+      */
+    case class Params(
+        nHeadPeers: Int,
+        nEvacs: Int,
+        payoutBatchSize: Int,
+    ) {
+        require(nHeadPeers >= 1, s"nHeadPeers must be >= 1, got $nHeadPeers")
+        require(nEvacs >= 0, s"nEvacs must be >= 0, got $nEvacs")
+        require(payoutBatchSize >= 1, s"payoutBatchSize must be >= 1, got $payoutBatchSize")
```
dragospe, line 39: We have non-negative and positive integer types we can use for these instead.

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/net/RBRNet.scala
```diff
@@ -0,0 +1,147 @@
...
+        require(
+          nEvacs % payoutBatchSize == 0,
+          s"nEvacs ($nEvacs) must be a multiple of payoutBatchSize ($payoutBatchSize)"
+        )
```
dragospe, line 43: Again, this is something addressed in ISO 15909-1 that gets very clunky with HLPN. Another point towards revision.

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/net/RBRNet.scala
dragospe, line 1: Let's be sure to add a json-compatible export function as a method.

Actually, it'd be even better to add a PNML export, but that might be too much of a lift to do from scratch.

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/net/Transitions.scala
```diff
@@ -1,7 +1,52 @@
...
+        // Consume + Restore across a synthetic TallyBufferPlace.
+        case TallyRestoreU
+        case TallyRestoreA
+        case TallyRestoreV
+        case TallyUU
+        case TallyUA
+        case TallyUV
+        case TallyAA
+        case TallyAV
+        case TallyVV
```
dragospe, line 48: Yikes.

integration/src/test/scala/hydrozoa/integration/rbr/model/RBRModel.scala
```diff
@@ -0,0 +1,64 @@
...
+  * A `MaxSteps` guard bounds the loop so a mistake in arc weights (or a future cyclic transition)
+  * does not lock a test up.
+  */
+object RBRModel {
+
+    /** Upper bound on the number of firings the driver will perform before giving up. Chosen so
+      * that a reasonable RBR run (e.g. 20 head peers, 1000 evacs, batch 63 ⇒ ~1050 firings)
+      * completes comfortably, while an unbounded cycle terminates the test loudly.
+      */
+    val MaxSteps: Int = 10_000
```
dragospe, line 23: I wonder how difficult it would be to solve the RBRNet for confluence. A formal proof of the net topology given constraints on the semantic would probably be easy-ish?

integration/src/test/scala/hydrozoa/integration/rbr/model/RBRModelTest.scala
```diff
@@ -0,0 +1,85 @@
...
+      VotedPlaceId -> 0,
+      AbstainPlaceId -> 0,
+      TallyBufferUnvotedPlaceId -> 0,
+      TallyBufferAbstainPlaceId -> 0,
+      TallyBufferVotedPlaceId -> 0,
+      PayoutObligationsPlaceId -> 0,
+      EvacuationOutputPlaceId -> params.nEvacs,
+      CollateralPlaceId -> params.nHeadPeers,
+      AmbientPlaceId -> 0,
+    )
```
dragospe, line 37: I don't love this as a Map. Perhaps it should just be a function?

dragospe: (Obviously this applies to the whole  interface of the classifier, for instance the same map in the evacuation tests.)

integration/src/test/scala/hydrozoa/integration/rbr/model/RBRModelTest.scala
```diff
@@ -0,0 +1,85 @@
...
+    private val scenarios: List[RBRNet.Params] = List(
+      RBRNet.Params(nHeadPeers = 1, nEvacs = 0, payoutBatchSize = 1),
+      RBRNet.Params(nHeadPeers = 1, nEvacs = 1, payoutBatchSize = 1),
+      RBRNet.Params(nHeadPeers = 3, nEvacs = 0, payoutBatchSize = 63),
+      RBRNet.Params(nHeadPeers = 3, nEvacs = 63, payoutBatchSize = 63),
+      RBRNet.Params(nHeadPeers = 3, nEvacs = 126, payoutBatchSize = 63),
+      RBRNet.Params(nHeadPeers = 3, nEvacs = 100, payoutBatchSize = 10),
+      RBRNet.Params(nHeadPeers = 7, nEvacs = 21, payoutBatchSize = 3),
+      RBRNet.Params(nHeadPeers = 20, nEvacs = 1_000, payoutBatchSize = 100),
+    )
```
dragospe, line 48: This should almost certainly be a generator instead. We can still tune for some of the corner cases you highlighted here, of course, rather than just uniform probability over the range.

integration/src/test/scala/hydrozoa/integration/rbr/model/RBRModelTest.scala
```diff
@@ -0,0 +1,85 @@
...
+
+    scenarios.foreach { params =>
+        test(s"terminal marking is analytical for $params") {
```
dragospe, line 51: "analytic" is weird phrasing here. I don''t think most people would know what that means without reading the source.

integration/src/test/scala/hydrozoa/integration/rbr/model/RBRModelTest.scala
```diff
@@ -0,0 +1,85 @@
...
+              s"\nexpected: ${expected(params)}\nactual:   $actual"
+            )
+        }
+    }
+
+    test("Params.require rejects nEvacs not divisible by payoutBatchSize") {
+        intercept[IllegalArgumentException] {
+            RBRNet.Params(nHeadPeers = 3, nEvacs = 100, payoutBatchSize = 63)
+        }
+    }
```
dragospe, line 64: Don't love this. 1 more point for a 15909 rewrite.

integration/src/test/scala/hydrozoa/integration/rbr/model/RBRModelTest.scala
```diff
@@ -0,0 +1,85 @@
...
+        intercept[IllegalArgumentException] {
+            RBRNet.Params(nHeadPeers = 0, nEvacs = 0, payoutBatchSize = 1)
+        }
+    }
+
+    test("Params.require rejects payoutBatchSize < 1") {
+        intercept[IllegalArgumentException] {
+            RBRNet.Params(nHeadPeers = 1, nEvacs = 0, payoutBatchSize = 0)
+        }
+    }
```
dragospe, line 76: These should just be type safe, see above.

integration/src/test/scala/hydrozoa/integration/rbr/property/Abstraction.scala
```diff
@@ -25,32 +25,32 @@ type RBRHistogram = Histogram[RBRPlaceId, Utxo]
...
-  * `collateral_input` of some downstream tx — rather than by a content sentinel. That
-  * requires shifting the classifier from `Utxo => Option[Bucket]` to a form that has
-  * access to the surrounding tx history.
+  * TODO: the `"collateral"` sentinel only exists in the synthetic `InitialDisputeUtxos` fixture —
+  * the real rule-based tx builders (`VoteTx`, `TallyTx`, `ResolutionTx`, `AbstainTx`,
+  * `RatchetVoteTx`, `EvacuationTx`) draw collateral from plain Ada wallet UTxOs whose `datumOption`
+  * is `None`, so content-based detection fails in end-to-end scenarios and collateral outputs land
+  * in [[AmbientPlaceId]]. The right fix is to bucket collateral by role in the tx graph — mark any
+  * output that is later consumed as a `collateral_input` of some downstream tx — rather than by a
+  * content sentinel. That requires shifting the classifier from `Utxo => Option[Bucket]` to a form
```
dragospe, line 38: This is an older comment, but I think the "right fix" is to actually just do 15909-1 style data nets with token identity (...or is that 15909-3?)

integration/src/test/scala/hydrozoa/integration/rbr/property/Abstraction.scala
```diff
@@ -25,32 +25,32 @@ type RBRHistogram = Histogram[RBRPlaceId, Utxo]
...
+  * that has access to the surrounding tx history.
   *
   * Usage: `Histogram.empty(RBRClassifier(using env)).addAll(utxos.map(Utxo(_, _)))`
   */
 class RBRClassifier(using env: MultiNodeConfig)
     extends Classifier[RBRPlaceId, Utxo](AmbientPlaceId):
```
dragospe, line 44: Random thought, but I wonder if there's any non-dependently-typed way to prove a function is a surjection. It's hard to tell just from the implementation whether this classification is "complete" in this way.

PR #541, opened 2026-07-20
Head API restructure, slice 2: timestamp + reverse-index capture writes

euonymos:
Slice 2 of the head-API effort ([slice 1: #540](https://github.com/cardano-hydrozoa/hydrozoa/pull/540)) — **capture writes only**, so the data the upcoming query endpoints need starts accumulating now. Independent of #540 (disjoint files, branched off main). No backward compatibility anywhere per design discussion — no active heads exist.

## Captures

- **Request receive time** — `UserRequestWithId.receivedAt: Instant`, stamped by the author peer in `RequestSequencer` at CR1 id assignment. Rides the peer wire (transport codec extended), so every peer's `Request` journal stores the author's stamp. Plain `Instant`, not `QuantizedInstant`: the request codec path deliberately carries no `CardanoNetwork.Section` and receive time is not slot-aligned data.
- **Soft-confirmation time** — new `Timestamped[P]` persistence wrapper; `FastConsensusActor.completeCell` stamps the `SoftConfirmation` write with the node-local saturation moment.
- **Hard-confirmation time** — `persistHardConfirmation` wraps the `HardConfirmation` value the same way.
- **`RequestBlockIndex`** — `(author, requestNum) → (blockNum, validity)`, one row per block event, written inside `JointLedger.snapshotBundleBatch`'s existing atomic bundle. Key is `[peer:4][requestNum:8]` big-endian, so a per-author listing is a contiguous prefix scan.
- **`BlockStackIndex`** — `blockNum → stackNum`, one row per block in the stack's `StackBrief` range, in the same atomic batch as the stack's `HardConfirmation`.

The confirmation timestamps are node-local observations (each peer's signature set saturates at its own moment) — diagnostic data, never consensus state. `receivedAt` is the only consensus-replicated stamp (single author).

## Readers updated

`RuleBasedActor`'s two R10-floor `HardConfirmation` gets, `HardConfirmationScan`, and the `StoreDump`/`JournalKey.decode` exhaustive matches over the sealed `Cf` (the latter two were caught by review as `-Werror`-only breakage — local compile passes, CI would not have). `design/persistence-and-crash-recovery.md` §3.3/§7 extended with the new CFs and value shapes.

## Verified

`just build-werror` green (all targets, CI flags); affected suites green: persistence (incl. new `StoreKeyTest` key-mapping + composite-key ordering cases), transport `CodecsTest`, `JointLedgerTest`, `ReplayActorTest`, `DisputeActorTest`. Stage suites run in CI.

## Next

Slice 3: read-model plumbing (a read-only consensus-store reader handed to `HydrozoaRoutes`, tapir path codecs for `HeadPeerNumber`/`RequestNumber`), then the `/head/blocks*` and `/head/requests*` endpoints over this data.


euonymos:

PR #542, opened 2026-07-20
Head API restructure, slice 3: consensus read model + /head/blocks endpoints

euonymos:
Slice 3 of the head-API effort, **stacked on #541** (base = `feature/head-api-capture`; retarget to `main` after #541 merges — this PR consumes its `Timestamped` values and `BlockStackIndex`).

## What

- **`ConsensusStoreReader`** — the read-only consensus-store view for the user-facing server, following the `EutxoL2LedgerReader` pattern: block-spine scan + point reads of briefs, soft/hard-confirmation records, and the block→stack index. Constructed from `Main`'s `Persistence` and threaded through `HydrozoaServer` into `HydrozoaRoutes`; `ConsensusStoreReader.empty` serves test/harness wiring.
- **`GET /head/blocks`** — number, round-robin leader (`blockNum % nHeadPeers`, pure config), type. Block 0 (the initial block) is synthesized from `HeadConfig` with no leader.
- **`GET /head/blocks/{block-number}`** — type, leader, and confirmation status `PROPOSED | SOFT | HARD` with this node's confirmation moments (ISO-8601), resolved via `SoftConfirmation` and `BlockStackIndex` → `HardConfirmation` from slice 2.
- **`GET /head/blocks/{block-number}/header`** — the header content (shape varies by block type); block 0 serves the config-derived `BlockHeader.Initial`.
- Validated tapir path codec for `BlockNumber` — malformed input is a 400 at decode, never a handler throw.

## Verified

`just build-werror` green (it caught value-discard warnings in the new test that the plain compile does not — fixed); new `HeadBlocksEndpointsTest` 5/5 through the HTTP layer (listing + leader math, the full PROPOSED→SOFT→HARD ladder with times, block-0 synthesis, header content, 404/400 edges); all server suites 20/20; core OpenAPI golden regenerated (second run green); `integration/Test` compiles.

Note: this branch does not include #540's path renames, so the regenerated `openapi.yaml` will conflict with #540's — whichever merges last regenerates the golden once (self-healing test).

## Next

Slices 4–5 remainder: `/head/requests` listing/per-peer/detail over the `Request` journals + `RequestBlockIndex` (needs the `HeadPeerNumber`/`RequestNumber` path codecs), then the per-block effects attribution.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -68,6 +68,76 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
+  /head/blocks:
+    get:
+      description: Every block of the head in block order — number, fast-cycle leader,
+        and type. Block 0 is the head's initial block (config, no leader).
+      operationId: getHeadBlocks
+      responses:
+        '200':
+          description: ''
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/BlockSummaryView'
+        default:
+          description: ''
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+  /head/blocks/{block-number}:
+    get:
+      description: 'One block''s type, leader, and confirmation status: PROPOSED,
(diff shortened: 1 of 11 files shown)


euonymos:

PR #543, opened 2026-07-20
Head API restructure, slices 4-5: /head/requests endpoints

euonymos:
Slices 4-5 of the head-API effort, **stacked on #542** (base = `feature/head-api-read`; retarget down the stack as #541 → #542 merge). Adds the request queries over the `Request` journals and the reverse indices from slice 2.

## What

- **`ConsensusStoreReader`** gains request reads: `requestsOf` (per-author `Request`-journal scan), `request` (point read), `requestBlock` (`RequestBlockIndex` point read).
- **`GET /head/requests`** — every request the head has assigned an id, across all head peers: object-form request id (`{headPeerNumber, requestNumber}`) + type (`transaction` / `deposit`).
- **`GET /head/requests/{head-peer-number}`** — one peer's requests, by number and type.
- **`GET /head/requests/{head-peer-number}/{request-number}`** — type, receive time (`receivedAt` from slice 2), and the lifecycle status: `UNPROCESSED | LOCALLY_PROCESSED (block + validity) | SOFT_CONFIRMED | HARD_CONFIRMED`, resolved through the `request → block` index and the block's confirmation records (reusing the blocks slice's confirmation-time helper). `relatedEffects` is always absent — per-request effect tracking (reverse-index #3) is its own ticket.
- **Validated tapir path codecs** for `HeadPeerNumber` (0..255) and `RequestNumber` (40-bit) — out-of-range input is a 400 at decode, never a throw in the types' `apply`.

Request ids are object-form everywhere in these responses, per the API decision; the packed i64 stays remote-ledger-only (the write-path `RequestAcceptedResponse` is unchanged in this slice).

## Verified

`just build-werror` green (caught the same `-Werror` value-discard class in the new test — fixed); new `HeadRequestsEndpointsTest` 5/5 through the HTTP layer (head-wide + per-peer listings, the full UNPROCESSED→HARD_CONFIRMED ladder with times and the invalid verdict, 404/400 edges); all server suites 25/25; core OpenAPI golden regenerated; `integration/Test` compiles.

## Remaining in the effort

The per-block effects endpoints (`/head/blocks/{n}/effects/*`) — per-block attribution over the per-stack `HardConfirmation` records — and eventually reverse-index #3 (request → effects) to fill `relatedEffects`.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -68,6 +68,86 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
+  /head/requests:
+    get:
+      description: Every request the head has assigned an id, across all head peers
+        — request id (author peer number + request number) and type (transaction or
+        deposit).
+      operationId: getHeadRequests
+      responses:
+        '200':
+          description: ''
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/RequestSummaryView'
+        default:
+          description: ''
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+  /head/requests/{head-peer-number}:
+    get:
+      description: Every request authored by one head peer, in request-number order
+        — request number and type.
+      operationId: getHeadPeerRequests
+      parameters:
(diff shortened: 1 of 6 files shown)


euonymos:

PR #544, opened 2026-07-20
Head API: reshape /head/requests (opaque id, filters, drop per-peer)

euonymos:
Applies the revised requests design (agreed in discussion) on top of the built requests slice (#543). Stacked on `feature/head-api-requests`; retarget down the stack as parents merge. The new block-effects subsystem is a separate follow-up PR (next).

## What changed vs #543

- **Listing rows** are now `{ requestId (opaque i64), peerNumber, requestType }` instead of the object-form id. The i64 is the packed `asI64` — the same value `POST /head/tx` returns in `RequestAcceptedResponse`, so `GET /head/requests/<that>` round-trips with the accept response.
- **Query filters** `?type=transaction|deposit` and `?peer_number=<n>` on the listing. An unknown `type` or out-of-range `peer_number` yields an empty list, not an error.
- **The per-peer path endpoint** `GET /head/requests/{head-peer-number}` is **removed** — `?peer_number=` replaces it.
- **Details** key by a single opaque segment: `GET /head/requests/{request-id}` (was `/{peer}/{number}`). The path codec bounds the i64 to `[0, 2^48)` so an out-of-range token is a clean **400** (`RequestId.fromI64` would otherwise throw an out-of-range peer/number and 500). The details body carries the opaque id, author peer, type, receive time, and lifecycle status.
- `relatedEffects` stays absent — request→effects remains a separate task.

Removes the now-dead `PeerRequestSummaryView` (+ mapper) and the `HeadPeerNumber`/`RequestNumber` path codecs (no path uses them anymore). OpenAPI golden regenerated.

## Decisions applied (from discussion)

Opaque id = **decimal i64 string**; requests reshape shipped **before** block-effects; `relatedEffects` deferred.

## Verified

`just build-werror` green; `HeadRequestsEndpointsTest` reshaped to the new surface (5/5: i64 rows, both filters incl. empty-result cases, the opaque-id lifecycle ladder, id/peer/receive-time/verdict in details, 404 for unknown id + 400 for malformed/out-of-range); all 25 server suites; golden regenerated; `integration/Test` compiles.

## Next

The block-effects read subsystem: `GET /head/blocks/{n}/effects` (per-block-type listing, decomposing per-stack partitions back onto blocks), the per-kind sub-resources incl. the SEC (head+coil signatures via a positional split at nHeadPeers, synthetic `blake2b_256(sec.header)` id), and `GET /head/effects/{l1TxId}` backed by a new reverse-index CF.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -70,10 +70,22 @@ paths:
                 $ref: '#/components/schemas/ErrorResponse'
   /head/requests:
     get:
-      description: Every request the head has assigned an id, across all head peers
-        — request id (author peer number + request number) and type (transaction or
-        deposit).
+      description: Requests the head has assigned an id — opaque request id, author
+        peer number, and type (transaction or deposit). Filter with ?type=transaction|deposit
+        and ?peer_number=<n>.
       operationId: getHeadRequests
+      parameters:
+      - name: type
+        in: query
+        required: false
+        schema:
+          type: string
+      - name: peer_number
+        in: query
+        required: false
+        schema:
+          type: integer
+          format: int32
       responses:
         '200':
           description: ''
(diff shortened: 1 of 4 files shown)


euonymos:

PR #552, opened 2026-07-22
feat(rbr): model the RBR dispute/evacuation flow as an HLPN

dragospe:
Third of three stacked PRs splitting the (now-closed) #538. **Stacked on #551** — review/merge that first.

`RBRHlNet` mirrors the L1 tx builders transition-by-transition — Vote, Abstain, RatchetVote, Tally (split by the `maxVote` winner), VotingDeadline, Resolution, Evacuation — on the `lib/petri` HLPN framework. Payout obligations are committed `TransactionOutput`s keyed by SEC version: Resolution reveals the winning version via a `ResolvedVersion` selector, and Evacuation drains only that version's obligations (losing SECs' commitments stay put), so the paid-out set follows the resolution outcome.

Verified: clean `integration` compile; **18 RBR tests pass**.

**Stack:** ① retire P/T (#550) → ② HLPN framework (#551) → ③ RBR net (this)


dragospe (commented):
## Design review: the `Output` color class narrows a color *domain* to *marking* data

Two concrete defects on `outputClass` (`RBRHlNet.scala:241-248`), and a modeling point they surface.

### The defects

1. **Fabricated fallback.** `committedObligations.headOption.map(_._2).getOrElse(committedOutputs(0).head)` (244) injects a version-0 output that corresponds to no real obligation — a phantom color in the domain — purely to satisfy `Sort.Class`'s `NonEmptySet`.
2. **Can't express the empty head.** A head with no committed outputs at fallback is unlikely but legal, and the model should express it. `NonEmptySet` forbids an empty carrier, which is *why* the fallback exists. Both defects are the same root cause seen twice.

### Root cause: the carrier is doing the marking's job

The net builder *manufactures* its output domain from `committedOutputs` (a fixture that distinguishes outputs by lovelace, `version*100 + j`, line 153), welding the net's type structure to one scenario. The CPN discipline is **fixed net, varying marking**: the color domain is fixed structure; the marking is the run. A scenario-dependent domain gives us *genuinely different net graphs per run* instead of one graph with different `M₀` — which is not what reachability/terminal-state analysis wants to compare.

Splitting "narrowing" three ways:

- **Size params (`nHeadPeers`, `maxVersionMinor`)** — legitimate net-*family* parameters. A 3-peer and 5-peer net genuinely differ. Keep.
- **Structural refinements (`validBallots`, `validOwners`)** — invariants masquerading as domains. The tell is at 249-254: the inscriptions are *already* the full products, reconciled to the narrow carrier by a ⊆ check (`domainCompatible`, `SortCheck.scala:62-67`). `Owner` is the clincher: it is **read-only** (only `b.read`, 388/416), so its diagonal-ness is *entirely* an `M₀` fact — narrowing its carrier encodes nothing the marking doesn't already determine.
- **Outputs** — opaque, only ever `Wildcard`-matched (606/644), never enumerated (`ModeSelector` treats `Wildcard` as match-only, 129/150). Model as an intensional/generic domain ("all outputs") and put the committed set in the marking. This **dissolves** the emptiness problem: the domain is infinite-hence-non-empty, and zero obligations is just an empty marking — no `NonEmptySet` violation, no fallback.

### Symmetry (for a future symbolic-reduction pass)

The usable symmetry group factorizes per class, so partial reduction on the symmetric subset is the *normal* mode, not a compromise. Notably:

- The inhibitor (644) is a **wildcard over `Output`** — a symmetric function — so it does **not** cost `Output` symmetry (SSR extends to inhibitor nets with symmetric inscriptions). Not the culprit.
- The real, inherent losses are the **ordered classes** (`Version`/`Status` → identity symmetry only, needed for ratchet/tally `<`) and the marking couplings (ownership diagonal restricts `Peer×Key` to the diagonal relabeling subgroup — still `Sₙ`).
- The **largest** available factor is `Sym(Output)` — currently destroyed by the lovelace-distinguishing carrier. The intensional-generic outputs change is the *prerequisite* that unlocks it.

### The refactor

- **Outputs → `Sort.Data`** (opaque, non-enumerable, ordered, may-be-empty/infinite domain); committed obligations passed in as a net argument seeding the marking. Deletes the fallback, fixes emptiness. *(this PR)*
- **`Owner` → `Peer×Key`**, diagonal moves to `M₀`. Read-only ⇒ zero new proof obligation. *(this PR)*
- **Ballots → full product**: **deferred** — trades the free `markingError` guard for explicit reachability-invariant tests, which must land first.
- **Size params**: unchanged.

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/hlpn/RBRHlNet.scala
```diff
@@ -0,0 +1,687 @@
...
+      * outputs (so a higher version carries a bigger evacuation batch), each distinguished by
+      * lovelace amount so tokens are distinct. Shared by the net's seed and tests.
+      */
+    def committedOutputs(versionMinor: Int): List[TransactionOutput] =
+        (1 to versionMinor.max(1)).toList.map { j =>
+            TransactionOutput(
+              payoutAddress,
+              Value(Coin((versionMinor.toLong * 100 + j) * 1_000_000L))
+            )
+        }
```
dragospe: This is wrong, double check

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/hlpn/RBRHlNet.scala
```diff
@@ -0,0 +1,687 @@
...
+      * `(i+1, (i+2 | 0 for the last, (Awaiting, 0)))`.
+      *
+      * `committedObligations` seeds `PayoutObligations` with the `(version, output)` payout tokens
+      * of each candidate SEC (see [[committedObligations]]), and `VotableVersions` with their
+      * distinct versions. `Output` is intensional, so an empty list is a valid, empty seed (a head
+      * with nothing to evacuate or vote on).
+      */
+    def apply(
+        nHeadPeers: Int,
+        committedObligations: List[(BigInt, TransactionOutput)],
```
dragospe: maybe a better representation is `Map[BigInt, List(TransactionOutput)]` (mapping minor version to the payout obligations)

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/hlpn/RBRHlNet.scala
```diff
@@ -0,0 +1,687 @@
...
+        // ---- Variables ----
+        val peer = Var("peer", peerClass) // the box owner (Vote / Abstain signer)
+        // the acting peer supplying collateral (ratchet / tally)
+        val collateralPeer = Var("collateralPeer", peerClass)
+        val key = Var("key", keyClass)
+        val link = Var("link", keyClass)
+        val status = Var("status", statusClass)
+        val version = Var("version", versionClass)
+        // the evacuated batch: all obligations of the resolved version, up to maxEvacuationsPerTx
+        val batch = CollectVar("batch", payoutSort, EvacuationTx.Assumptions.maxEvacuationsPerTx)
```
dragospe, line 241: Do we still need this one?

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/hlpn/RBRHlNet.scala
```diff
@@ -0,0 +1,687 @@
...
+              Value(Coin((versionMinor.toLong * 100 + j) * 1_000_000L))
+            )
+        }
+
+    /** The scenario's committed obligations across candidate SEC versions `1..maxVersionMinor`,
+      * keyed by version — the standard `PayoutObligations` seed passed to [[apply]], shared with
+      * tests. An empty list (no candidate versions) is a valid, empty seed.
+      */
+    def committedObligations(maxVersionMinor: Int): List[(BigInt, TransactionOutput)] =
+        (1 to maxVersionMinor).toList.flatMap(v => committedOutputs(v).map(BigInt(v) -> _))
```
dragospe: I think this is dead code, check and remove

integration/src/test/scala/hydrozoa/integration/rbr/model/petri/hlpn/RBRHlNet.scala
```diff
@@ -0,0 +1,687 @@
...
+                ballots <- b.place(Ballots, RBRPlace(initialBallots, ballotSort))
+                owner <- b.place(Owner, RBRPlace(ownership, ownerSort))
+                votingOpen <- b.place(VotingOpen, RBRPlace(oneDot, Sort.Dot))
+                votingClosed <- b.place(VotingClosed, RBRPlace(noDots, Sort.Dot))
+                unresolvedTreasury <- b.place(UnresolvedTreasury, RBRPlace(oneDot, Sort.Dot))
+                resolvedTreasury <- b.place(ResolvedTreasury, RBRPlace(noDots, Sort.Dot))
+                regimeRef <- b.place(RegimeRef, RBRPlace(oneDot, Sort.Dot))
+                disputeScriptRef <- b.place(DisputeScriptRef, RBRPlace(oneDot, Sort.Dot))
+                treasuryScriptRef <- b.place(TreasuryScriptRef, RBRPlace(oneDot, Sort.Dot))
+                setupLadder <- b.place(SetupLadder, RBRPlace(oneDot, Sort.Dot))
```
dragospe: Possibly should be starting with 7? So we can choose which to use?

PR #567, opened 2026-07-28
fix(config): encode CardanoNetwork.Custom protocolMagic correctly

rssh:
## What

The `Custom`-network JSON encoder wrote `cardanoInfo.protocolParams` under the `protocolMagic` key,
while the decoder reads a `Long` there — so a `Custom` network never round-tripped and the magic was
silently dropped. This emits the actual `protocolMagic`.

## Test

Adds `CardanoNetworkCodecTest`:

- the three standard networks round-trip through JSON;
- a `Custom` network encodes `protocolMagic` as the magic number (the exact regression).

## Impact / scope

Latent today — no `Custom` network is constructed anywhere yet — so there is no runtime impact; this
unblocks upcoming custom/devnet support.

Note: a *full* `Custom` decode additionally requires the `ProtocolParams` / `CardanoInfo` codec to
round-trip, which is a separate, deeper issue (the scalus blockfrost `ReadWriter` is write/read
asymmetric). This PR is the `CardanoNetwork`-level half.

Generated by Claude Opus 4.8 (1M context)

### DIFF
--- a/src/main/scala/hydrozoa/config/head/network/CardanoNetwork.scala
+++ b/src/main/scala/hydrozoa/config/head/network/CardanoNetwork.scala
@@ -86,7 +86,7 @@ object CardanoNetwork {
             case CardanoNetwork.Custom(cardanoInfo, protocolMagic) =>
                 Json.obj(
                   "custom" -> cardanoInfo.asJson,
-                  "protocolMagic" -> cardanoInfo.protocolParams.asJson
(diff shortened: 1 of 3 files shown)


rssh:
Superseded by #570, which now targets main and includes this encoder fix (commits 701da862, 0dc51e9c) alongside the full custom-network feature (WI-001..005 + e2e + Yaci deploy path).

PR #568, opened 2026-07-28
fix(config): round-trip ProtocolParams/CardanoInfo JSON for Custom networks

rssh:
## What & why

A `Custom` (e.g. Yaci devnet) head-config must serialize its `CardanoInfo` and read it back on `serve`, but the `ProtocolParams` / `CardanoInfo` circe codec did not round-trip:

- hydrozoa's wrapper was internally mismatched — the encoder emitted a JSON **object** while the decoder expected a JSON **string**.
- scalus itself exposes no exact round-tripping JSON codec for `ProtocolParams`: `blockfrostParamsReadWriter` is write/read asymmetric (its `cost_models` writer emits arrays, the reader expects objects → upickle `TraceException`), and `cardanoCliParamsReadWriter` routes `UnitInterval` / `NonNegativeInterval` through `Double`, which is lossy (a mainnet `priceSteps` of `0.0000721` decodes back as `0.000072`).

This blocked Phase 0b/1 custom-network work — a `Custom` head-config could not fully deserialize.

## What changed

`Codecs.scala` now derives a **structural, human-observable** circe codec for `ProtocolParams`, preserving every field exactly:

- `UnitInterval` / `NonNegativeInterval` → exact `{ "numerator": <Long>, "denominator": <Long> }` (shared `fraction*` helper; `Try`-wrapped so the decoders stay total, since both constructors `require` a positive denominator)
- `cost_models` → id-keyed object (`"0"`, `"1"`, `"2"`)
- `ProtocolVersion` decoder is likewise `Try`-wrapped (it `require`s `major >= 1`), keeping the whole `ProtocolParams` decode total
- everything else (`ExUnits`, `ExUnitPrices`, both voting-thresholds, `ProtocolParams`) derived from the record structure

This makes `ProtocolParams` consistent with its already-derived container/siblings (`CardanoInfo`, `Network`, `SlotConfig`, `Utxo`) rather than being the lone opaque-string exception.

## Tests

`CardanoNetworkCodecTest` asserts a direct `ProtocolParams` round-trip and a full `Custom`-network round-trip over the real mainnet/preprod/preview params — strong fixtures, since their tiny `ExUnitPrices` expose any precision loss — plus a regression test that a malformed `ProtocolVersion` decodes to a `Left` rather than throwing. All pass; `CI=true sbt Test/compile` (`-Werror`) clean.

## Review notes

Ran `/simplify` (deduped the interval codecs into the shared `fraction*` helper; flattened `costModelsDecoder`) and `/code-review`. The review **refuted the main risk** — the encoding-format change can't break existing consumers, because the old codec was already broken (encoder emitted an object, decoder expected a string), so no working old-format `Custom` config exists, and standard networks encode as bare strings. Its robustness findings are addressed here: total `ProtocolVersion` decoder, leaf-before-aggregate codec ordering, and private-helper placement.

**Known limitation (pre-existing, low severity):** the derived decoders don't honor case-class default parameters, so a hand-authored `Custom` config that *omits* a defaulted `SlotConfig` field (`epochLength`, `zeroEpoch`) fails to decode rather than falling back to the default. Round-trips are unaffected (the encoder always emits all fields); closing this would mean switching to configured decoders file-wide.

---

Delivers anthill **WI-005**. Stacked on #567 (the `Custom` protocolMagic encoder fix it builds on).

https://claude.ai/code/session_019kwekvQLBqvm8n9oXaU194

### DIFF
--- a/src/main/scala/hydrozoa/lib/cardano/scalus/codecs/json/Codecs.scala
+++ b/src/main/scala/hydrozoa/lib/cardano/scalus/codecs/json/Codecs.scala
@@ -5,7 +5,7 @@ import io.circe.generic.semiauto.{deriveDecoder, deriveEncoder}
 import io.circe.{Decoder, DecodingFailure, Encoder, Json, KeyDecoder, KeyEncoder, parser}
 import scala.util.Try
 import scalus.cardano.address.Network
-import scalus.cardano.ledger.{CardanoInfo, KeepRaw, ProtocolParams, SlotConfig, Transaction, TransactionHash, TransactionInput, TransactionOutput, Utxo}
+import scalus.cardano.ledger.{CardanoInfo, CostModels, DRepVotingThresholds, ExUnitPrices, ExUnits, KeepRaw, NonNegativeInterval, PoolVotingThresholds, ProtocolParams, ProtocolVersion, SlotConfig, Transaction, TransactionHash, TransactionInput, TransactionOutput, UnitInterval, Utxo}
 import scalus.crypto.ed25519.SigningKey
 import scalus.uplc.builtin.ByteString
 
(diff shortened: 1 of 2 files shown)


src/main/scala/hydrozoa/lib/cardano/scalus/codecs/json/Codecs.scala
```diff
@@ -5,33 +5,98 @@ import io.circe.generic.semiauto.{deriveDecoder, deriveEncoder}
...
+      * while the reader expects objects), and `cardanoCliParamsReadWriter` routes the
+      * `UnitInterval`/`NonNegativeInterval` fields through `Double`, which is lossy (a mainnet
+      * `priceSteps` of `0.0000721` decodes back as `0.000072`). We instead derive circe codecs
+      * field-by-field from the record structure and serialize the two fraction types as exact
+      * `{ "numerator": <Long>, "denominator": <Long> }` objects, so a `Custom` head-config's
+      * `CardanoInfo` stays readable and survives a serialize/deserialize cycle unchanged.
+      *
+      * The leaf codecs are defined before the derived aggregates that summon them; the two private
+      * fraction helpers they share sit at the foot of the object, per the file's ordering rule.
+      */
```
dragospe, line 29: perhaps this can be upstreamed too?

rssh: Yes, we fixed this in upcoming 1.0.0-M1  
 (so, this patch will be obsolete, if we move to the next scalus version,  from other side, -M1 is not release and it is unclear when 1.0.0 final will be published)

PR #570, opened 2026-07-29
feat: custom (Yaci-devnet) network support — serve, CLI, and a Docker smoke-test

rssh:
## What & why

First-class support for running a Hydrozoa head against an L1 endpoint other than the public
Blockfrost standard networks: a local **Yaci DevKit** devnet (a chain not known offline), or one of
the standard chains served from a **private Blockfrost-compatible endpoint** (e.g. a self-hosted
node). This enables local, hermetic, repeatable head development and is the L1 backing for a new
black-box e2e propagation test.

Previously serve + the config CLI were hardwired to the public standard networks: two `???`
backend-URL stubs on the `Custom` path, `keygen-fleet` / `deploy-scripts-and-g2-setup` /
`build-head-config` each derived the network from the Blockfrost key prefix and rejected anything
else, and the `CardanoNetwork.Custom` JSON codec didn't round-trip.

## What's in it

**Config + codecs**

- `CardanoNetwork.Custom` JSON now round-trips — the encoder wrote `cardanoInfo.protocolParams`
  under `protocolMagic`. Adds `ProtocolParams` / `SlotConfig` / `CardanoInfo` codecs, with the
  `Network` and `ProtocolVersion` decoders wrapped so an out-of-range byte is a `DecodingFailure`
  rather than a thrown constructor `require`.
- `NodePrivateConfig.cardanoBackendUrl` (optional) holds the node's backend URL.
  `CardanoBackendBlockfrost.apply(network, cardanoBackendUrl, apiKey, tracer)` folds the
  network→backend decision into one place — removing the two `???` stubs and the duplicated network
  `match` in `HeadConfig` / `NodeConfig`.
- A URL-backed network builds its `BlockfrostProvider` from the already-resolved `CardanoInfo`
  (params / slot config / magic are never re-fetched, one fewer HTTP round-trip); backend URLs are
  trailing-slash-normalized so sub-clients never emit `//`.

**How a network is sourced — `BootstrapNetwork`**

Two arms that mirror where a network's `CardanoInfo` comes from:

- `Known(network: StandardCardanoNetwork, backendUrl: Option[String])` — one of the three standard
  chains, fully known offline from its baked-in `CardanoInfo`; the optional `backendUrl` serves it
  from a private Blockfrost-compatible endpoint while keeping that baked-in info. This closes the
  "standard network + custom Blockfrost URL" gap: routing such a chain through the genesis-fetch
  path would mis-derive a Byron-prefixed chain's slot config (mainnet/preprod slots off by the whole
  Byron era → every L1 validity interval wrong).
- `Custom(blockfrostUrl, yaciAdminUrl)` — a chain not known offline (a Yaci devnet), resolved
  **live** at `build-head-config` time by `CustomNetworkResolver`: the Yaci admin API for a devnet
  whose slot `startTime` moves on each reset, else the generic `/genesis`. A bare `Custom` that
  resolves to a standard magic is rejected, pointing at the explicit `Known` path.

`keygen-fleet` records the decision in `defaults.json`; both `build-head-config` and
`deploy-scripts-and-g2-setup` read it through the shared `Bootstrap.readBootstrapNetwork` /
`resolveNetwork`, so the network is decided once (deploy-scripts' parallel `networkOfBlockfrostKey`
derivation is gone). The documented deploy procedure is unchanged — same flags, same steps.

**CLI (generation ladder)**

- `init-bootstrap-files`: `--cardano-network` accepts `custom`, plus new `--blockfrost-url` /
  `--yaci-admin-url`.
- `keygen-fleet`: derives the network from the private template — explicit `cardanoNetwork` name
  (with `cardanoBackendUrl` as its private-endpoint override), or `"custom"` / a bare
  `cardanoBackendUrl` for a live-resolved chain, else the key prefix — and writes the marker.
- `deploy-scripts-and-g2-setup` and `build-head-config` gain `--blockfrost-url` /
  `--yaci-admin-url`, overriding the in-mesh marker so a host-side build reaches Yaci at its
  host-mapped port while the containers keep the in-mesh URL.
- `peer-private.template.json` gains `cardanoBackendUrl` / `yaciAdminUrl` (both `null` by default).

**Docker e2e propagation test** (`DockerPropagationTest`, heavy, CI-excluded)

- Stands up 4 head-peer containers on a local Yaci devnet, runs the DEPLOYMENT.md sequence (keygen →
  topup → deploy-scripts → build-head-config → compose up), submits an L2 tx to head-0, and asserts
  it propagates to all four peers' L2 ledgers; tears down + dumps container logs on failure.
- `SubmitL2Transaction.buildTx` / `parseUtxoView` widened to `private[hydrozoa]` so the suite builds
  the same tx it submits over HTTP.
- Run via `just integration-e2e-docker`. Excluded from CI via an env-conditional `Tests.Exclude`
  (which also repairs the pre-existing, silently-0-tests `integration-yaci` recipe). Documented in
  `docs/integration-stages.md`.

**Deployment**

- `docker-compose.yml` gains an optional `yaci` profile; DEPLOYMENT.md §7 documents running a head
  against a local Yaci DevKit alongside the preprod/Blockfrost path.

## Validation

- `CI=true sbt Test/compile integration/Test/compile` (`-Werror`), scalafmt, scalafix — all clean.
- Config/codec round-trip tests: a `Custom`-network `HeadConfig` round-trip, a `CardanoNetwork`
  codec suite, and a `BootstrapNetwork` round-trip in `BootstrapMembershipTest`.
- Live runs below were performed **before** the final `BootstrapNetwork` (Known/Custom) refactor,
  which reworked network resolution and touched the e2e suite — they should be re-run before merge:
  - **Yaci path** — `DockerPropagationTest` green end-to-end (submit → propagate to all 4 peers,
    self-tears-down).
  - **Preprod path** — brought up live on real preprod: head formed (init tx confirmed on-chain), an
    L2 tx propagated across peers, then finalize → recover funds → teardown.

https://claude.ai/code/session_01YNjANrzpw9BtwMuF1oKq4h


rssh:
About docker-compose in main and integration tests:  they are different; integration tests have only yaci-devkit, and it is in tests and not intended to be one of the artifacts used externally.  Main docker-compose is what external users can use to start (and it have yaci-devkit profile)

docs/integration-stages.md
```diff
@@ -274,6 +274,25 @@ The default 2-peer run takes seconds; 10-peer WS runs at `commits=500` take a fe
 
 ---
 
+## E2E: 4-node Docker propagation (CI-excluded)
+
+`integration/src/test/scala/hydrozoa/integration/e2e/DockerPropagationTest.scala` is a third, **black-box** level that sits outside the `ModelBasedSuite` framework. Instead of wiring actors in-process, it stands up **four real head-peer containers** on a local **Yaci** devnet via `docker compose`, forms a head through the shipped artifacts (the packaged image, the mesh, the user HTTP API), submits an L2 transaction to head-0 over HTTP, and asserts it propagates to every peer's L2 ledger.
+
+- **What it exercises that the stages can't.** The actual deployment path end to end: `docker compose`, four distinct node identities reaching consensus over the real WS mesh, the packaged image, and the custom-network (Yaci) config flow (`keygen-fleet` → `topup` → `deploy-scripts-and-g2-setup` → `build-head-config` → `serve`). Stage 1/4 are in-process and never touch the image, compose, or the HTTP surface.
```
euonymos: Let's check - I remember Peter added a mechanism to check _some_ HTTP server components.

docs/integration-stages.md
```diff
@@ -274,6 +274,25 @@ The default 2-peer run takes seconds; 10-peer WS runs at `commits=500` take a fe
 
 ---
 
+## E2E: 4-node Docker propagation (CI-excluded)
+
+`integration/src/test/scala/hydrozoa/integration/e2e/DockerPropagationTest.scala` is a third, **black-box** level that sits outside the `ModelBasedSuite` framework. Instead of wiring actors in-process, it stands up **four real head-peer containers** on a local **Yaci** devnet via `docker compose`, forms a head through the shipped artifacts (the packaged image, the mesh, the user HTTP API), submits an L2 transaction to head-0 over HTTP, and asserts it propagates to every peer's L2 ledger.
```
euonymos: Do we want to run coil peers as well?

docs/spec/integration-stages.md
```diff
@@ -274,6 +274,25 @@ The default 2-peer run takes seconds; 10-peer WS runs at `commits=500` take a fe
...
+- **How it drives bring-up.** It shells out to the staged `hydrozoa` launcher and `docker compose`, and reuses the existing Yaci admin client (`integration/.../yaci/DevKit.scala`) for `topup`/`devnetInfo`, plus the server DTOs (`ApiDto`, `SubmissionClient`) for submit + polling. Convergence — the same utxo/feed entry appearing on all four peers — *is* the propagation assertion; on failure it dumps each container's logs.
+- **Why it's CI-excluded.** It is minutes-long and needs Docker + a Yaci container + the built image. Its FQN is in `Tests.Exclude(...)` in `build.sbt` (survives even `integration/testOnly *`), exactly like `Stage1PropertiesYaci`.
+- **How to run.**
+
+```bash
+just integration-e2e-docker   # builds the image (Docker/publishLocal), stages the launcher, runs the suite
+# equivalently, after `just docker-image` + `just stage`:
+sbt "integration/testOnly hydrozoa.integration.e2e.DockerPropagationTest"
+```
```
euonymos, line 289: Let's add just's target - it's better than copy-pasting from docs.

docs/integration-stages.md
```diff
@@ -274,6 +274,25 @@ The default 2-peer run takes seconds; 10-peer WS runs at `commits=500` take a fe
 
 ---
 
+## E2E: 4-node Docker propagation (CI-excluded)
```
euonymos: I think it should be called "Docker smoke-test" - I am not fully sure yet, but I guess we check not only propagation.

integration/src/test/resources/e2e/docker-compose.yaci.yml
```diff
@@ -0,0 +1,100 @@
+# A 4-head-peer Hydrozoa head for the e2e L2-propagation test (DockerPropagationTest): 4 head peers, each a single
+# hydrozoa node running the built-in EUTXO L2 ledger in-process, all publishing their user/admin
+# HTTP API (host 8080-8083). (The production docker-compose.yml is a different shape: 2 head + 4
+# coil on Blockfrost.io.)
+#
+# Two modes:
+#   - preprod (default):  `docker compose -f … up`
```
euonymos: It's not clear why a file called 'docker-compose.yaci.yml' is talking about preprod.

integration/src/test/resources/e2e/docker-compose.yaci.yml
euonymos, line 1: I think we need to use the same docker-compose.yml - one of the ideas of that test is to test out deployment procedure is sound - if we use a separate file, we won't catch those errors.

src/main/resources/scaffold/peer-private.template.json
```diff
@@ -22,6 +22,8 @@
     }
   },
   "blockfrostApiKey": "preview00000000000000000000000000000000",
+  "cardanoBackendUrl": null,
```
euonymos: This should be called `blockfrostApiUrl`.

src/main/resources/scaffold/peer-private.template.json
```diff
@@ -22,6 +22,8 @@
...
+  "yaciAdminUrl": null,
```
euonymos: Why do we need it? To instantiate the slotConfig?
I don't want to add this to the config - Yaci is purely a testing tool; no code in main should be aware of it. 

Instead, I guess, we should fix slotConfig/protocolParams directly in the config - the test harness prepares all that and fill in the missing fields.

src/main/scala/hydrozoa/app/DeployScriptsAndG2Setup.scala
euonymos, line 1: The idea is right - let's define the network explicitly so we don't have to use the key as a carrier. But it should be done in a way that is transparent: no Yaci mentions, should work with any blockfrost-compatible backend, and should know nothing about which one is used.

src/main/scala/hydrozoa/bootstrap/Bootstrap.scala
euonymos, line 1: Same thing here.

src/main/scala/hydrozoa/multisig/backend/cardano/CustomNetworkResolver.scala
euonymos, line 1: Again, doesn't fit the idea of test-tools ignorant main codebase.

build.sbt
```diff
@@ -229,12 +229,23 @@ lazy val integration: Project = (project in file("integration"))
...
+      // HYDROZOA_INCLUDE_HEAVY_TESTS=1 is set — that is how the `just` recipes run one by FQN:
+      //   HYDROZOA_INCLUDE_HEAVY_TESTS=1 sbt "integration/testOnly …Stage1PropertiesYaci"   (just integration-yaci)
+      //   HYDROZOA_INCLUDE_HEAVY_TESTS=1 sbt "integration/testOnly …DockerPropagationTest"  (just integration-e2e-docker)
+      Test / testOptions ++=
+          (if (sys.env.get("HYDROZOA_INCLUDE_HEAVY_TESTS").contains("1")) Seq.empty
+           else
+               Seq(
+                 Tests.Exclude(
+                   Seq(
+                     "hydrozoa.integration.stage1.Stage1PropertiesYaci",
```
euonymos, line 288: Do we really want to exclude stage1?

DEPLOYMENT.md
```diff
@@ -564,3 +569,94 @@ The finalization tx pays the L2 state and equity back out on L1 (watch the netwo
...
+| Who | Store / Blockfrost API | Admin API |
+|---|---|---|
+| **containers** (in-mesh) | `http://yaci:8080/api/v1` | — |
+| **host** (CLI generation steps) | `http://localhost:18080/api/v1` | `http://localhost:10000/local-cluster/api` |
+
+So the peers' `cardanoBackendUrl` is the **in-mesh** URL, while the host-side `deploy-scripts` and
+`build-head-config` are given the **host** URLs via `--blockfrost-url` / `--yaci-admin-url` (which
+override the in-mesh URL that `keygen-fleet` wrote into `defaults.json`).
+
+### Step 1 — start Yaci and create the devnet
```
euonymos: The first step should be a link to the Yaci installation.

DEPLOYMENT.md
```diff
@@ -564,3 +569,94 @@ The finalization tx pays the L2 state and equity back out on L1 (watch the netwo
...
+
+```bash
```
euonymos: That drudgery should be automated with scripts.

DEPLOYMENT.md
```diff
@@ -564,3 +569,94 @@ The finalization tx pays the L2 state and equity back out on L1 (watch the netwo
...
+Yaci has no faucet; top up head-0's address straight through the admin API:
+
+```bash
+ADDR=$(just head-zero-address)                          # head peer 0's bech32 address
+curl -X POST http://localhost:10000/local-cluster/api/addresses/topup \
+     -H 'Content-Type: application/json' \
+     -d "{\"address\":\"$ADDR\",\"adaAmount\":100000}"  # bump if build-head-config reports a shortfall
+```
+
+### Step 4 — deploy scripts + build the head config, pointed at Yaci
```
euonymos: Steps should be preserved; we don't want to describe them twice.
The main section should cover the yaci-based setup.

The difference with Yaci should be eliminated where possible:
 - no access to Yaci admin; all that should be done by the Yaci harness that spins up, waits for the initialization, and gives back slot config and everything else, which are just constants for public testnets
 - maybe before that, we need to generate the keys - topping up might be a part of that "spin up yaci procedure"
 - everything else likely should be exactly the same

justfile
```diff
@@ -195,7 +195,15 @@ integration:
...
-  sbt "integration/testOnly hydrozoa.integration.stage1.Stage1PropertiesYaci"
+  HYDROZOA_INCLUDE_HEAVY_TESTS=1 sbt "integration/testOnly hydrozoa.integration.stage1.Stage1PropertiesYaci"
+
+# Heavy 4-node Docker e2e L2-propagation test (Yaci devnet): builds the image, stages the launcher,
+# then runs the CI-excluded DockerPropagationTest. Needs a running Docker; minutes-long. See the
+# E2E section of docs/integration-stages.md.
+integration-e2e-docker:
+  #!/usr/bin/env bash
+  trap 'just notify "integration-e2e-docker"' EXIT
+  HYDROZOA_INCLUDE_HEAVY_TESTS=1 sbt Docker/publishLocal stage "integration/testOnly hydrozoa.integration.e2e.DockerPropagationTest"
```
euonymos: Good, we already have it. Please ignore my previous comment - or even better, mention the just command, not the sbt commands in readme.

euonymos (changes requested):

euonymos:
> Custom(blockfrostUrl, yaciAdminUrl)

- this is the point we need to reconsider.

euonymos:
> About docker-compose in main and integration tests: they are different; integration tests have only yaci-devkit, and it is in tests and not intended to be one of the artifacts used externally. Main docker-compose is what external users can use to start (and it have yaci-devkit profile)

As I mentioned in the review, we do want to test exactly the same docker-compose in this test on purpose - one of the strengths of this test is that it may guarantee that our (not some different test-specific) deployment procedure is correct.

euonymos:
Good gap to close, but the layering's the blocker, and it's behind most of my comments: main must know nothing about Yaci and do no network resolution — it reads a complete config (slot config + params + magic), and the test harness resolves those from Yaci and writes them. So CustomNetworkResolver / yaciAdminUrl / --yaci-admin-url all move out of main.      
                                                                                              Keep the codec fix + `CardanoNetworkCodecTest`, the `Known(net, url)` path, and              `rejectStandardMagic` — those are generic and good.                                           
                                                                                              
e2e: reuse the real docker-compose.yml via an override, not a fork.                         
                                                                                              
Ignore my build.sbt stage1 comment — I misread, stage 1 was excluded before your PR.

PR #572, opened 2026-07-30
fix(l2-remote): reuse one WebSocket connection; fail soft when unreachable

euonymos:
RemoteL2Ledger opened a fresh `JdkWSClient` (a new JDK `HttpClient` — a new selector thread + file descriptors) **and** a new connection on **every request**, then retried failures infinitely.

Under load this exhausts the process fd limit (`Too many open files`); the JDK wraps that `SocketException` in a fatal `java.lang.InternalError`, which cats-effect does not catch through `handleErrorWith`, so the node crashes.

## Changes
- **One shared `WSClient` + one connection**, opened lazily and reused across all requests, serialized by a `Mutex` so concurrent callers never interleave frames; reopened once transparently if it drops. `create` now returns a `Resource` owning the client and closing the connection on shutdown.
- **Soft failure:** when unreachable, requests return `L2LedgerError("remote L2 ledger unavailable, try again later")` through the normal `EitherT` error channel — no infinite retry, no crash. Application-level response errors leave the connection intact.
- **Request timeout** on each exchange so a silent remote cannot wedge the now-shared connection (default 30s).
- Replaced the `ConnectionError`/`ReconnectionAttemptFailed` events with a single `Unavailable` (warn).

Passes `just fmt`, `just lint`, and the full `-Werror` build (main + test + integration).

### DIFF
--- a/src/main/scala/hydrozoa/app/Serve.scala
+++ b/src/main/scala/hydrozoa/app/Serve.scala
@@ -221,12 +221,10 @@ object Serve {
                 )
                 for {
                     _ <- Resource.eval(log.info(s"L2 ledger: remote at $wsUri"))
-                    ledger <- Resource.eval(
-                      RemoteL2Ledger.create(
-                        wsUri = wsUri,
-                        config = nodeConfig,
-                        tracer = tracer,
-                      )
+                    ledger <- RemoteL2Ledger.create(
+                      wsUri = wsUri,
+                      config = nodeConfig,
+                      tracer = tracer,
                     )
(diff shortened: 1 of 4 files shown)


euonymos:
Closing — superseded by the L2 connectivity redesign. The uniform fail-soft policy here was right for screening but wrong for the mutation path (a consensus-agreed command must land, not be dropped); the redesign splits screening from mutations and makes the mutation path retry-through-safe via a command-number-monotonicity contract. The shared-HttpClient fd-leak fix is carried into that work.

PR #574, opened 2026-07-31
chore: bump Scalus 0.18.1 -> 1.0.0

rssh:
Bumps `scalusVersion` to 1.0.0 (library, `scalus-cardano-ledger`, `scalus-testkit`, and the compiler plugin — all published for Scala 3.3.7).

**No Scala source changes were needed for the upgrade itself.**

## Script hashes change

1.0.0 makes protocol version 11 (van Rossem) the default compile and evaluation target, so both rule-based validators recompile. `src/main/resources/plutus.json` is regenerated via `runMain hydrozoa.rulebased.ledger.l1.script.plutus.Export`:

| Validator | hash | size |
|---|---|---|
| Dispute Resolution | `da13a4db…` → `4c5dd09b…` | 12091 → 10618 B (−12.2%) |
| Rule-Based Treasury | `95f006ce…` → `0a7a58a7…` | 9138 → 8084 B (−11.5%) |

⚠️ **Anything already deployed at the old hashes must be redeployed.** That is inherent to the PV11 switch; reproducing the old output would mean pinning `Options.plomin`.

## Migration-guide items checked against this codebase

Against [migrating-to-1.0](https://scalus.org/docs/get-started/migrating-to-1.0):

- **Removed API** (`ReprTag`, `ScalusDebug`, `CompileDerivations`, `w7l`, `bloxbean.TxEvaluator`, `multiIndexArray`) — no usages here.
- **`BigInt./` and `%` switched from floor to truncated division** — the only on-chain uses are in `Scalar.scala`, where every operand is provably non-negative (the `[0, fieldPrime)` field invariant, plus the `if e < 0` guard in `scale`), so the two semantics coincide.
- **Hand-written `Eq` instances now rejected on-chain** — the sole instance is `given Eq[VoteStatus] = Eq.derived`, which is the sanctioned form.
- **`ImmutableEmulator`/`Emulator` gained constructor params** — all defaulted; test call sites are unaffected.

## Also in this PR

- `CLAUDE.md` said Scala 3.3.6; `ThisBuild / scalaVersion` has been 3.3.7.
- `UtxoToPlutusBenchmark.run()` bound its only computation to an unused local and declared `Unit`, making it a dead-code-elimination candidate — the benchmark may have been timing nothing. It now returns its `TxInInfo`, matching the other three `run()` methods in that file. This also clears the one `-Werror` failure in `benchmark`, which pre-dates this branch (`git blame`: 2025-09-17) and only surfaced because CI does not compile that subproject.

## Verification

| Check | Result |
|---|---|
| `Test/compile` + `integration/Test/compile` + `benchmark/Test/compile`, with `-Werror` | clean |
| `sbt test` | 366/367 |
| `integration/testOnly *` (CI's fast set) | 39/39 |
| `fmtCheckAll` + `lintCheckAll` | clean |

The single unit-test failure is ``PropertyM.`factor` cli utility works``, which shells out to GNU `factor`; that binary comes from the Nix stdenv and was absent from the shell I ran in. It is environmental and unrelated — CI runs under `nix develop`.

All on-chain suites pass under PV11, including the two ScalaCheck property suites (`Dispute Vote Attack`, `RBR Dispute Resolution Commands`) that actually evaluate the compiled scripts in the emulator.

## Known follow-up (not addressed)

`HashUtxoBenchmark.run()` has the same dead-code hazard as the benchmark fixed here, hidden behind a `: Unit` ascription rather than tripping the warning. Fixing it changes what that benchmark measures, so it is left alone.

https://claude.ai/code/session_01YNjANrzpw9BtwMuF1oKq4h


euonymos (approved):
LGTM. Minimal, well-scoped bump with a thorough migration audit.

Independently verified the one item with real semantic risk — the BigInt `/`/`%` floor→truncated change. Across the whole `cardano-onchain` (`@Compile`) subproject the only division sites are in `Scalar.scala`; the two `% fieldPrime` run on non-negative sums/products (the `[0, fieldPrime)` field invariant), and the `/` in `scale` is guarded by `if e < 0 then zero`, so floor and truncated coincide everywhere. The other `@Compile` files have no division. Claim is both correct and complete.

Script-hash change → redeploy is inherent to the PV11 switch and correctly flagged; deploy/bootstrap configs pinning the old hashes will need regenerating. Benchmark `run()` returning `TxInInfo` is the right JMH fix (and clears a real -Werror failure).

Follow-up (not this PR): CI doesn't compile the `benchmark` subproject — worth adding so warnings there stop rotting.

PR #575, opened 2026-08-03
test(mbt): drive RBR dispute→evacuation MBT

dragospe:
Stacked on #571 (`hlpn/5-yaci-backend`); review that first — this branch adds the MBT↔Yaci wiring on top, then builds the RBR fallback→evacuation MBT up to full model fidelity (L2 transactions, trailing minors, withdrawals). Also carries `main`'s **scalus 0.18.1 → 1.0.0** bump (merged via `hlpn/4-rbr-mbt`), which is what makes the Yaci image's cost models correct (see the scalus ExUnits bug below).

Parameterizes `RbrMbtSuite` so a single suite class can run against either the in-memory `CardanoBackendMock` or a real Testcontainers-managed Yaci devnet, selected by `backendSpec: BackendSpec = Mock | Yaci(YaciConfig)`.

## Status: passes end-to-end on both backends

The RBR dispute → evacuation now runs setup → deposits → **L2 transactions** → fallback → the full rule-based dispute → evacuation, and the terminal L1 snapshot projects to the same `ObservableMarking` as the model net (`alpha == beta`) on **both** the Mock and a real Yaci devnet (3 head + 3 coil peers). Zero `sutErrors`, zero build failures.

The earlier *"premature done / under-evacuation (28/45 outputs)"* blocker is **resolved**: `alpha` is now sourced from the head's own committed evacuation-map size, observed live from `StackComposer.CommittedMap` peer traces (`N`, keyed on the last on-chain-settled major `M`), instead of a fixed obligation count — so the model tracks exactly what the head resolves to under fallback, including trailing minors and hard-confirmed-but-rolled-back majors. Racing deposits that never commit are absent from `N`, matching what lands on L1, so stragglers need no separate committed-set accounting.

## What the MBT exercises (model fidelity)

- **Setup phase = stage4's generator.** The pre-fallback phase (`RbrMbtScenarioGen`) delegates to a shared, SUT-generic `SetupScenarioGen` factored out of stage4's own scenario generator — same Poisson superposition, same L2-tx / deposit / absorption command mix. Any new setup path stage4 grows flows into the RBR setup phase for free; the two differ only by a small `Config` (datum sentinels, deposit validity, output addresses).
- **L2 transactions + trailing minors.** The setup phase submits real L2 txs (reusing stage4's `genL2TxCommand`), producing minor blocks. The Poisson inter-arrival timing lets the SUT settle a major mid-scenario, so L2 txs land as **trailing minors** on majors ≥ 1 — which exercises the `RuleBasedActor`'s trailing-minor `defaultVoteMap` resolution in an integration test (previously believed unhittable here). Verified `N` resolving to e.g. the `(2,1)`/`(4,2)` minor SEC.
- **Withdrawals (first-class).** L2 txs run stage4's full strategy mix including `RandomWithdrawals`. A withdrawal (CIP67 flag-1 output) exits to L1 and is accounted **separately** from evacuations: a distinct `"withdrawal"` datum sentinel, an `RBRClassifier` bucket, and an inert `WithdrawalOutput` place in `RBRHlNet`. The committed `EvacuationMap` (the `N` trace) is flag-2 only — withdrawals are separate payout obligations (`JointLedger`), so `EvacuationOutput` stays uncontaminated by timing; `W` is observed from the firewall across every cleared tx (settlement, rollout, and the post-fallback evacuation of pending withdrawals), deduped by input. Verified: Mock `EvacuationOutput 60==60` / `WithdrawalOutput 11==11`, Yaci `51==51` / `2==2`.

## How the Yaci run got unblocked (product fixes)

Driving the MBT against a real devnet turned up a chain of issues; each is fixed in this branch. **The original "Bug A (firewall × coils wedge)" and "Bug B (index-7 off-by-one)" framings were both wrong** — the wedge was the 120s takeoff crossing the 60s WS idle timeout, and the "index-7 nonexistent input" was the expected winner/loser race (`BadInputsUTxO` on inputs a peer already spent), not an off-by-one.

Fixes, in the order the flow unblocked them:

1. **`PPViewHashesDontMatch` on every Plutus tx** — the `0.10.0-preview2` Yaci image ships a broken cost-model bootstrap. Resolved by the scalus 1.0.0 bump, which defaults to `yaci-cli:0.12.0-beta5` (proper cost models).
2. **`ExUnitsTooBigUTxO` on the evacuation tx** — a scalus bug (see next section); worked around in `EvacuationTx.Build`.
3. **`TreasuryEvacuationRedeemerParseError: "not a constructor"`** stalling evacuation after batch 1 — `CardanoBackendBlockfrost` fetched the spending redeemer by `datum_hash` instead of `redeemer_data_hash`. **Fixed** → `getRedeemerDataHash`.
4. **CardanoLiaison handoff storm** (~thousands of `FallbackToRuleBasedDispatched`/run) — `reconcileTargetState` re-announced the handoff every tick. **Debounced**: re-announce only when the observed rule-based treasury utxo changes.
5. **Store-hammering poll cadence** (test-side) — `yaciTxTiming` bumps deposit maturity so the CL poll can slow to 1s (`poll ≤ depositMaturity/5`); the `submitKickRequest` loop stops once fallback dispatches.
6. **Fatal `Plutus script evaluation failed` on the evacuation build** — the actor read the treasury utxo (`getTreasury`) and the evacuation map / withdrawal history from *separate* store projections that lag independently on Yaci, so a peer could assemble an `EvacuationTx` from an internally-incoherent snapshot. **Fixed** by folding both reads into one: `lastContinuingTxs` returns each continuing output as a resolvable `Utxo` (`CardanoBackend.ContinuingTx`), and `Evacuation.handle` derives the treasury utxo *and* the remaining map from the same chain read (`loadEvacuationState`). A stale-but-coherent chain now fails at submission (recoverable `BadInputsUTxO`) instead of fatally at build time, restoring the "build failure == real bug" invariant.

## What's here (test scaffolding)

- **`RbrMbtSuite` backend param + env ADT.** `Env` goes from `Unit` to a sealed `RbrMbtEnv = Mock | Yaci(DevKit)`, threaded `initEnv` → `genInitialState`. Mock path is byte-equivalent to before.
- **`YaciDevnet.acquireShared`** — JVM-wide `YaciContainer` singleton (no release hook), since `ModelBasedSuite.initEnv` returns a bare `Env`.
- **Live protocol params.** `yaciNetwork` fetches the devnet's params (incl. cost models) from Blockfrost instead of the bundled `yaci-params.json`, so Plutus txs don't fail `PPViewHashesDontMatch`.
- **No `reset()` in the Yaci path.** `reset()` wedges the yaci-store on the pre-reset tip (topups never confirm); the single MBT iteration runs against the pristine container instead.
- **`RbrMbtPropertiesYaci`** runner (3+3, `Tests.Exclude`-gated; run via `just integration-yaci-docker`).

## ⚠️ Upstream scalus bug: `Ordering[ExUnits]` is memory-first → `ExUnitsTooBigValidator` under-validates

scalus 1.0.0's `Ordering[ExUnits]` (`scalus/cardano/ledger/Types.scala`) is lexicographic, **memory-first**:

```scala
given Ordering[ExUnits] = (x, y) =>
    if x.memory != y.memory then x.memory.compareTo(y.memory) else x.steps.compareTo(y.steps)
```

`scalus.cardano.ledger.rules.ExUnitsTooBigValidator` decides "over the per-tx budget?" with `actualTxExecutionUnits > maxTxExecutionUnits` using that ordering — so it is governed by **memory alone** whenever the memories differ. A tx that is over on CPU **steps** but **under** on memory compares as *not* over budget, so the validator passes it — but the real ledger checks each component independently and rejects it (`ExUnitsTooBigUTxO`). The correct check is component-wise: `actual.memory > max.memory || actual.steps > max.steps`.

This silently under-validates **any** high-CPU/low-memory tx, not just evacuation. Minimal reproduction (also committed as `ExUnitsOrderingScalusBugTest`, kept green by asserting the buggy behaviour):

```scala
import scala.math.Ordered.orderingToOrdered
import scalus.cardano.ledger.ExUnits

val max    = ExUnits(memory = 16_500_000L, steps = 10_000_000_000L)  // per-tx budget
val actual = ExUnits(memory =  6_966_471L, steps = 11_606_514_781L)  // under mem, OVER steps

assert(actual.steps > max.steps)   // true  — the tx IS over budget (on CPU steps)
assert(actual > max)               // FAILS — the Ordering says it is NOT over budget
```

`EvacuationTx.Build` works around it by re-checking ex-units component-wise before accepting a built tx (halving otherwise). Remove that workaround and this test once upstream is fixed.

## Remaining / follow-ups

- **Refund verification — blocked on fund14 (deferred).** A deposit that isn't absorbed before fallback should reappear as a refund output on L1, but **refund-tx L1 submission is unimplemented**: `partitionRefunds` builds the `RefundTx`s into `PartitionEffects`, but nothing submits them (`CardanoLiaison.handleStackL1Effects` skips them, `StackEffectsBuilder` TODO fund14). So no refund output reaches L1 today and an `alpha == beta` refund check would be vacuous. The ready design — mirror `WithdrawalOutput`: a `"refund"` sentinel via `genRegisterDepositCommand`'s `refundDatum` (arbitrary, stamped verbatim on the single refund output/deposit), a `RefundOutput` place/classifier, firewall-observed `R` — is documented in the `beforeFinalize` TODO. A small task once fund14 lands.
- **Stop hard/soft-confirming stacks once in RBR.** The CL handoff debounce removes the re-announce spam, but the head keeps hard-confirming stacks after fallback (~36/s), each driving an ungated `runEffects` poll. The proper fix is to stop confirming stacks once the head is in the rule-based regime.

## Verification

- `just build-werror` (scalus 1.0.0) + scalafmt/scalafix clean.
- `RbrMbtProperties` (Mock, 3+3, L2 txs + withdrawals) passes: `alpha == beta` on `EvacuationOutput`/`WithdrawalOutput`, `N` resolving to a trailing-minor SEC.
- `RbrMbtPropertiesYaci` (real devnet, 3+3) passes end-to-end with zero `sutErrors` / build failures: `EvacuationOutput 51==51`, `WithdrawalOutput 2==2`.
- RBR/petri unit suites (`RBRHlNetTest`, `ObservableMarkingTest`, `EvacuationPropertyTest`, `CommitmentSelectionPropertyTest`, …) and `ExUnitsOrderingScalusBugTest` green.


dragospe:
### Progress update — the Yaci MBT now passes end-to-end (`OK, passed`)

The **premature-done under-evacuation** blocker called out in "Remaining / follow-ups" is resolved. `RbrMbtPropertiesYaci` now runs setup → deposits → fallback → rule-based dispute → **full evacuation** with `alpha == beta` and zero `sutErrors`. Landed since the last body update:

- **Rollback-aware evacuation candidate set** (`8fac3107`) — the RBA now resolves to the latest hard-confirmed SEC whose major matches the fallback treasury's major `M`, covering the full resolvable set (`votableSecs` + `(M,0)` default-vote base). Fixes the `UnknownResolvedKzg` crash and the "peers conclude done while payouts are missing" under-completion.
- **Confluent tally fold + N-agnostic `driveToEvacuated`** (`09ff085a`).
- **Deposit-evacuation path now exercised** (`7162f1b5`, `0d9f6717`) — deposits were always taking the refund path (`committed-deposits=0`) because stage4's 2-min deposit validity outlasts the 90s commit window; RBR now uses 20s validity and routes deposit outputs to peer-owned addresses. Deposits absorb into majors and commit on-chain before fallback.

### New: observe the committed L2 map size from peer traces (`3bccefb0`)

Toward collapsing most of `ModelState`: the model can read the committed evacuation-map size the head resolves to **directly from the peers** instead of reconstructing it from deposit accounting.

- Added a production `StackComposer.CommittedMap(version, size)` trace, emitted per committed block as the running `EvacuationMap` is persisted.
- The harness computes `N` = size of the max-minor `CommittedMap` at the **last on-chain-settled major `M`** (`M = max` of settlements that cleared the firewall). Keying on `M` — not the max committed major — excludes the hard-confirmed-but-unsettled fallback majors that run *ahead* of on-chain settlement (the slow cycle hard-confirms majors the dropped fallback settlement never lands).
- **Verified (deposits-only, non-breaking diagnostic):** across runs `N == obligationCount == firstPayoutsLeft` (e.g. `committedMaps 27→33→42` over majors 0..4 with `M=2` → `N=42`; another run `N=47`). `alpha` still uses `obligationCount` for now.

**Next:** flip `alpha` to `N`, delete the deposit committed-set accounting (most of the deposit bookkeeping in `ModelState` goes away), then wire L2 transactions + stage4's `genDelayForAbsorption` — where the `M`-keyed resolution starts to matter (L2 minor blocks change committed map sizes above `M`).

integration/src/test/scala/hydrozoa/integration/rbr/mbt/NetDriver.scala
```diff
@@ -1,85 +1,45 @@
...
+  * head's autonomous dispute: every peer votes, the voting deadline passes, the ballot-box linked
+  * list folds pairwise to the single fully-tallied `(0, 0)` box, resolution reveals the winning
+  * version, and evacuation drains its whole committed batch.
+  *
+  * Fully driven by the [[ModeSelector.unifying]] simulator and parametric in the peer count. The
+  * tally's `continuing.key < removed.key` fold guard makes the contraction confluent — key 0 (the
+  * public box) is never removed, so it always survives to the `(0, 0)` terminal — which means
+  * firing each phase to saturation reaches the same marking regardless of `nHeadPeers`, the vote
+  * pattern, or the fold order the selector happens to pick. `ObservableMarking` projects the
+  * winning version away, so any resolved version yields the same observable terminal.
```
dragospe: This comment is redundant; remove it

src/test/scala/hydrozoa/multisig/backend/cardano/FirewalledCardanoBackend.scala
```diff
@@ -2,11 +2,10 @@ package hydrozoa.multisig.backend.cardano
 
 import cats.effect.IO
 import hydrozoa.lib.logging.ContraTracer
-import hydrozoa.multisig.backend.cardano.CardanoBackend.Error
+import hydrozoa.multisig.backend.cardano.CardanoBackend.{ContinuingTx, Error}
```
euonymos, line 5: Arguably this whole file (FirewalledCardanoBackend.scala + event) should be in test codebase.

euonymos (approved):

PR #577, opened 2026-08-04
perf(test): share derived peer keys and wallets across TestPeers instances

rssh:
## What & why

`sbt test` took **38:57** on my machine, almost all of it re-deriving BIP32 keys rather than running tests. This brings it to **~14 min**.

A sampling profile (10 thread dumps of the forked test JVM) put **9 of 10 samples** in `CIP1852.getKeyPairFromMnemonic`, under `pbkdf2HmacSha512`.

Two compounding causes:

- **Caching the Bloxbean `Account` does not cache the expensive part.** `Account.publicKeyBytes()` re-runs the whole derivation on every call (`getHdKeyPairFromDerivationPath` → `getRootKeyPairFromMnemonic` → `pbkdf2HmacSha512`), and `hdKeyPair()` likewise. So `TestPeers.accountCache` memoized the cheap object and none of the cost.
- **`TestPeers.headPeers` was a `def`** that derives every peer's key, and `HeadPeers.Section` computes `headPeerNums` / `headPeerIds` / `headPeerVKeys` / `nHeadPeers` from it as `final def`s. A single `nHeadPeers` call inside a ScalaCheck generator therefore cost N full derivations — repeated per sample, across every property in the suite.

## What's in it

- `headPeers` becomes a `lazy val`.
- The derived key and wallet move to caches in the **companion object**. Companion, not instance: `TestPeers.arbitrary` builds a fresh instance per ScalaCheck sample, so an instance-scoped cache — as the existing `addressCache` / `walletCache` were — still pays a full derivation per peer per sample. `TrieMap`, since suites run in parallel and now share them.
- `addressCache` deliberately stays instance-scoped: unlike the key it is derived from, an address *is* network-specific.

## The assumption, and the test that pins it

The caches are keyed by `(mnemonic, ordinal)` and **not** by network, which is only sound because BIP32 derivation does not involve the network — only address encoding does, downstream of the key. The key space is therefore tiny: the seed phrases in use × at most `TestPeerName.maxPeers`.

A new property in `TestPeersTest` pins that. It derives through Bloxbean **directly** rather than via `TestPeers` — routing it through the caches it exists to justify would make it vacuously true — and compares both the public key bytes and, via `PeerWallet`'s extensional equality (same exported vkey *and* same signature over a fixed message), signing behaviour. Passes at 500 samples.

## Validation

Measured on `hydrozoa.config.head.initialization.InitializationParametersTest` — a property suite
sitting directly on the `TestPeers` generator path — run on each branch with a warm compile.
**CPU time (user+sys, including the forked test JVM), not wall clock**: this is a shared
workstation, and wall clock there measures contention as much as it measures work.

| | wall | CPU | minus sbt's 42s startup floor |
|---|---|---|---|
| `main` | 4m53s | 5m35s | **293s** |
| this branch | 0m17s | 1m07s | **25s** |

**~12x less CPU** for the suite itself, 5x end-to-end. Both runs executed the same 2 properties,
both passing — the caches return the same values the uncached path did.

The whole suite improves in the same direction, but I am not quoting a figure for it: the
before/after runs I have were taken under different machine load, so the number would say as much
about what else was running as about this change.

Supporting evidence, independent of timing: a sampling profile of the test JVM on `main` (thread
dumps, so a composition measure that load cannot distort) puts the top frame in
`CIP1852.getKeyPairFromMnemonic` under `pbkdf2HmacSha512`, reached via
`TestPeers.verificationKeyFor` -> `Account.publicKeyBytes`. After the key cache, the residual
derivation moves to `TestPeers.walletFor` exactly as predicted — hence the second cache — and the
top frame becomes `scalus...plutus.prelude.List.map2`, i.e. actual Plutus evaluation.

384 tests pass, up one: the new property.

https://claude.ai/code/session_01B21A9azBmPNgQjRrxGnN1r

### DIFF
--- a/src/test/scala/test/TestPeers.scala
+++ b/src/test/scala/test/TestPeers.scala
@@ -19,6 +19,7 @@ import org.http4s.Uri
 import org.scalacheck.Arbitrary.arbitrary
 import org.scalacheck.Test.Parameters
 import org.scalacheck.{Gen, Prop, Properties}
+import scala.collection.concurrent.TrieMap
 import scala.collection.immutable.SortedMap
 import scala.collection.mutable
 import scalus.cardano.address.ShelleyAddress
@@ -73,7 +74,10 @@ case class TestPeers private (
     // API
     // ===================================
 
-    override def headPeers: HeadPeers = {
+    // A `lazy val`, not a `def`: [[HeadPeers.Section]] derives headPeerNums / headPeerIds /
+    // headPeerVKeys / nHeadPeers from this, and generators call those per sample — recomputing the
+    // peer set each time meant re-deriving every peer's key (see [[verificationKeyFor]]).
+    override lazy val headPeers: HeadPeers = {
         def helper[A](f: TestPeerName => A) =
             NonEmptyList.fromListUnsafe(
               peerNumbers.map(ix => f(TestPeerName.fromOrdinal(ix)))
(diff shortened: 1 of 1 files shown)


euonymos:
Lol, accounts' caching gave me a significant improvement when I added it - this is why I decided it did the job.

PR #582, opened 2026-08-05
test(e2e): 4-head propagation, on a topology of its own

rssh:
Stacked on #570 — it needs `EutxoL2QueryClient` and `scripts/yaci-devnet.sh` from there. Based on `feat/custom-network-serve` so the diff shows only this change; GitHub will retarget to `main` when #570 merges.

## Why a second Docker suite

`DockerSmokeTest` (in #570) runs the head we ship — 2 head peers, 4 coils — and has to keep doing so: its value is that a failure implicates DEPLOYMENT.md. But only head peers publish the HTTP API (`runCoilNode` starts no `HydrozoaServer`), so it can assert propagation across **two** peers and no more. A fan-out that drops the last recipient, or a quorum off-by-one, would not show up.

This suite buys four HTTP-observable peers by giving up the "same as production" property — and says so in its name rather than quietly reshaping the smoke test's topology. Until #570, this 4-head shape *was* the only Docker test, wearing the shipped topology's clothes and carrying its own compose fork to make that work. Separating them is the point.

## What's here

- `Docker400PropagationTest` — `DockerHeadSuite` bound to a 4-head, 0-coil topology, under its own compose project.
- `integration/src/test/resources/e2e/docker-compose.4-0-0.yml` — named HEADS-COILS-QUORUM, matching `keygen-fleet`'s argument order.
- CI exclusion by FQN in `build.sbt`, plus `just integration-e2e-docker-400`.

## Why quorum 0

Nothing can reach a higher one. The head multisig is *all* head peers plus `MOf(coilQuorum, coilPeerVKeys)` (`HeadConfig.scala:530-539`) — head peers sign unanimously, there is no m-of-n over them — and `HeadConfig` refuses a quorum above the coil count. With no coil peers, 0 is the only accepted value. So coil fan-out is **not** covered here; that is the smoke test's job. The two are complementary, not redundant.

## Caveats

- **Not yet run.** It compiles and the devnet script beneath it is verified end to end, but no head has formed in containers on this topology yet.
- **Cannot run concurrently with `DockerSmokeTest`** — both reach peer *i* at `localhost:${8080 + i}` and both publish the devnet's 18080/10000. Different project names isolate containers and networks, not host ports.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -242,7 +242,8 @@ lazy val integration: Project = (project in file("integration"))
                  Tests.Exclude(
                    Seq(
                      "hydrozoa.integration.stage1.Stage1PropertiesYaci",
-                     "hydrozoa.integration.e2e.DockerSmokeTest"
+                     "hydrozoa.integration.e2e.DockerSmokeTest",
+                     "hydrozoa.integration.e2e.Docker400PropagationTest"
                    )
                  )
                )),
--- a/integration/src/test/resources/e2e/docker-compose.4-0-0.yml
+++ b/integration/src/test/resources/e2e/docker-compose.4-0-0.yml
@@ -1,0 +1,75 @@
+# A 4-head-peer, 0-coil-peer head — the `keygen-fleet 4 0 0` topology, hence the file name
+# (HEADS-COILS-QUORUM, matching the argument order). Used only by Docker400PropagationTest.
+#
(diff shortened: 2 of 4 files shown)


rssh:
Closing — reopening later once #570 lands. The branch stays as `test/docker-4-0-0-propagation`.

PR #620, opened 2026-08-15
feat: standalone `evacuate` command for the rule-based regime

euonymos:
## Summary

Adds a standalone **`hydrozoa evacuate`** command that starts the rule-based regime (dispute + evacuation) from a left-over database plus the same head + private config `serve` takes — the tool for recovering funds from a head whose process is gone. Along the way it fixes three real correctness bugs in the rule-based flow (tx signing, SEC signature-list alignment, and the evacuation candidate-map set) and simplifies the wallet model, logging, and Docker deployment.

## What's in it (commit by commit)

- **read-only store open** — `RocksDbBackendStore.openReadOnly` + `Persistence.fromBackendReadOnly` (reads generation/anchors without writing). The rule-based regime only reads persistence, so the standalone reader opens a stopped head's store read-only.
- **injectable CardanoLiaison notification** — the liaison's "switch to rule-based" signal becomes a callback; the multisig manager fires the handoff as before, a rule-based-only owner passes a no-op.
- **`RuleBasedRegimeManager` can own a `CardanoLiaison`** — `None` on the multisig→rule-based handoff (the multisig liaison stays alive for rollback resilience); `Some` on the standalone path so the fallback gets submitted and rollouts finish.
- **the `evacuate` subcommand** — boots only the L1 boundary (no consensus/mesh/HTTP/L2), store opened read-only, resident; `--data-dir` points it at a stopped head's store.
- **sign with `ownWallet`, retire the separate rule-based wallet** — dispute ballot boxes are locked to each peer's consensus key by the fallback tx, so every rule-based tx must be signed by `ownWallet`; the actor was signing with a distinct, unfunded `ruleBasedWallet`, so tally failed with `MissingVKeyWitness`. Sign everything with `ownWallet` and remove `ruleBasedWallet` from the config schema, keygen, template, and tests (one funded consensus key covers the whole flow — the only config ever exercised end-to-end).
- **store SEC signatures sparse and coil-peer-aligned** — the on-chain `DisputeResolutionScript` verifies coil header signatures **positionally** against `regimeDatum.coilPeers`, so a signing coil at position *i* must land at slot *i*. The SECs were carrying coil signatures **densely** (a packed prefix), so whenever the signing coils weren't a `[0..q)` prefix the vote tx failed script evaluation with `VoteCoilSigInvalid`. Store `headerMultiSigned` as an ordered `List[Option[HeaderSignature]]` aligned to `allHeadPeers.sorted ++ allCoilPeers.sorted` (`None` for non-signers); `HardAckAggregator.collectSecSignatures` keeps the `None` holes (map, not flatMap), `SlowConsensusActor` supplies the full `allSecPeers` order, and `RuleBasedActor.loadAction` splits at `nHeadPeers` into dense head `signatures = head.flatten` + sparse `coilSignatures`. Regression guard: the `multisignHeaderSparse` test fixture signs a **non-prefix** subset of coils (which a dense encoding would fail).
- **cover votable minor SECs in the evacuation candidate maps** — `loadEvacuationInputs` only assembled `candidateEvacMaps` in two terminal branches of the fallback-anchor walk (a Regular stack carrying a fallback → base + votable SEC maps; the Initial stack → initial map only). A head that fell into rule-based with just an Initial stack plus minor-only Regular stacks (one L2 tx, then the fallback fired, before any major settled) terminated on the Initial branch and loaded only the initial map — but voting (`votableSecs`) can resolve the dispute onto one of those minor SECs, leaving evacuation with no preimage and looping on `UnknownResolvedKzg`. Derive the candidate set the same way voting selects — `defaultVoteMap` + every `votableSecs` map (`loadCandidateEvacMaps`), independent of the anchor — and use the walk only to locate the fallback tx (`loadFallbackAnchor`). New `EvacuationInputsTest` reproduces the gap (RED before, GREEN after).
- **surface Plutus eval logs in VoteTx build errors** — the `VoteTx` build error now renders the underlying `EvaluationFailed(cause).logs` instead of an opaque `toString`, so a failing dispute-resolution script prints its Plutus trace.
- **size `minAdaForTreasury` to the real treasury utxo** — `maxNoLiabilitiesTreasuryUtxoBytes` was 6960 (a wildly high assumption); the actual no-liabilities treasury utxo is ~204 bytes, so it drops to 400. Defaults regenerated.
- **logging** — route `RuleBasedActor` / `CardanoLiaison` at info in `logback-docker.xml` (they inherited root=warn), drop the normal submit error to debug, raise the per-tick retry heartbeat to info.
- **Docker/deploy** — default images to `:latest` (still pinnable via `HYDROZOA_VERSION`; local `just docker-image` also tags `:latest`); fold the durable-state + `evacuate` profile into the single scaffolded compose (state is durable now); add `head-up` / `head-down` / `evacuate` / `evacuate-peer` / `head-reset` just recipes.
- **docs** — new `docs/spec/evacuate-command.md` + deployment guide updates.

## Testing

- `-Werror` build green (main + integration); all other modules test-compile.
- Full core unit suite passes; `DisputeActorTest`, `RuleBasedActorCoilTest`, `EvacuationInputsTest`, bootstrap/keygen/codec tests green.
- Validated live on preview against an `evacuate` container end-to-end: the head enters the rule-based regime, resolves the dispute onto a minor SEC, and drains the L2 map to L1. A `head0→head1 5 ADA` L2 transfer is reflected in the evacuated outputs (head1 receives 11+5 = 16 ADA, head0 5 ADA). This run also exercised the two signing/alignment fixes above (tally `MissingVKeyWitness` and the coil `VoteCoilSigInvalid`).

## Not in scope / follow-ups

- A dedicated `evacuate` **integration test** (seed a store → assert fallback → tally → drain) is not yet built; rule-based behavior is covered by the existing `integration/.../rbr/` suites plus the new `EvacuationInputsTest` unit.
- `evacuate` needs an image that contains it — `:latest` once a release ships this branch, or a local `just docker-image` until then.
- The crash-recovery "serve detects rule-based regime on boot" handoff (persistence spec §8 step 7) reuses this same `RuleBasedRegimeManager`; wiring that is a separate change.


src/main/scala/hydrozoa/app/Evacuate.scala
```diff
@@ -0,0 +1,161 @@
...
+            persistence <- Resource.eval {
+                given CardanoNetwork.Section = nodeConfig
+                // Read-only: the store was opened read-only, so this must not bump the arrival-stamp
+                // generation or record a zero-time (both are writes). The rule-based regime only
+                // reads (design `docs/spec/evacuate-command.md`).
+                Persistence.fromBackendReadOnly(backendStore, persistenceTracer)
+            }
+
+            system <- ActorSystem[IO]("Hydrozoa Evacuate")
+        } yield (backend, nodeConfig, labelNum, persistence, system)
```
dragospe, line 139: Perhaps we define a HydrozoaAppM or HydrozoaApp object that abstracts all of this setup?

Not blocking for this PR, but my gut says:
- We're going to write more of these commands, either for production or testing
- They're all going to be substantially similar in shape

src/main/scala/hydrozoa/rulebased/RuleBasedRegimeManager.scala
```diff
@@ -1,38 +1,95 @@
...
+  *     `RuleBasedActor` — `ownLiaison = None`.
+  *   - the standalone **`evacuate`** command, where no multisig `CardanoLiaison` pre-exists. There
+  *     `ownLiaison` is `Some(...)`, and this manager also spawns a `CardanoLiaison` so the fallback
+  *     gets submitted and in-flight rollouts finish. That liaison re-recovers its full target state
+  *     from persistence at boot; its "switch to rule-based" notification is a no-op (there is no
+  *     regime manager to hand off to — the `RuleBasedActor` running alongside picks the regime up
+  *     directly off L1).
+  *
+  * The actors themselves own every persistence read and every chain interaction; this shell exists
+  * only so the parent can spawn a single "regime manager" child per handoff.
```
dragospe, line 28: It might be simpler to just have the RBRM always spawn its own cardano liasion. 

What do you think?

dragospe: What I'm thinking is that we're currently saying "I have a cardano liasion for the multisig regime, but there's a bunch of stuff that I _don't_ want it to do in the rule-based regime. Let's add flags to take all those out"; whereas I would think "Let's make a baseline cardano liasion and compose it with the new behaviors we need in the multisig regime"; subtractive vs additive/compositional.

dragospe (commented):
Good catch on the dense signing. 

Regarding the rest, its plausibly correct in behavior, but I think I would structure things different and try to abstract some bits. Nothing blocking for a merge, but lets make a ticket if you agree.

src/main/scala/hydrozoa/rulebased/RuleBasedRegimeManager.scala
euonymos: Yeah, I think so. No reason to keep both managers; [Ticket](https://linear.app/gummiworm-labs/issue/GUM-274/always-run-cardanoliaison-in-the-rule-based-regime).

src/main/scala/hydrozoa/app/Evacuate.scala
euonymos: [Ticket](https://linear.app/gummiworm-labs/issue/GUM-275/factor-out-hydrozoaapp-object-that-abstracts-common-setup-for-commands)

PR #622, opened 2026-08-17
Reject L2 reports that break solvency or value conservation

dragospe:
## Problem

A remote L2 ledger can report evacuation-map diffs that mint or destroy value — the head would absorb the report and eventually multisign an on-chain commitment its treasury can't (or over-)cover. The concrete trigger was a sugar-rush-ledger rounding error: at 2 USDr / 1 ADA, a 1-lovelace sale debited the buyer 1 micro-USDr but credited the maker 2, inflating the L2's books one token unit per trade with nothing crossing the L1 boundary.

## Design

The head's double-entry balance identity

```
treasury.value == evacuationMap.totalValue + equity + beacon    (exactly, per asset)
```

decomposes into an **anchor** plus two **inductive steps**. Each leg is guarded where its data lives, and only the leg facing an untrusted party is a runtime check:

1. **Map side — runtime conservation gate** (`StackEffectsBuilder.checkPartitionConservation`): each L2 command's reported diff group must move the map by exactly what that command moved across the L1 boundary — a transaction by minus its payouts, the deposit-decisions group by the absorbed deposits' `l2Value` — in the coin and in every asset, with a block-level aggregate backstop. Per command, not per block: compensating per-tx errors cancel in aggregates. Rejections (`Error.EvacuationMapNotConserved`) carry the originating `RequestId` and fire before any effect tx is built or signed. The deposit gate is likewise tightened to equality: spawned L2 outputs must equal `depositL2Value` exactly, at screening and registration.

2. **Treasury side — property test** (`SettlementTxSeqBuilderTest`): the settlement builder's exactness is derivable from the tx balancer + equity arithmetic — our own code, so per "derivable ⇒ property," it's established by a randomized prop rather than re-checked at runtime: the produced treasury's liability pool (value − equity) moves by exactly absorbed − payouts across generated settlements (up to 100 payouts, rollout chains included, token dimension covered).

3. **Anchor — parse at the boundary**: the identity is verified once wherever a (treasury, map) pair enters the system: `InitializationTx.Parse` rejects an unbalanced treasury output (`TreasuryNotBalanced`) at config load, and `StackComposer.State.recover` refuses to boot from a store whose independent treasury/map snapshots diverge (store corruption, like a missing key).

An earlier revision of this branch enforced leg 2+3 as an O(map) runtime re-check at every commitment point (`checkTreasuryBalanced`, commit 2); it is removed in favor of the above in commit 4 — see the commit message for the reasoning.

## Commits

1. **Add double-entry accounting invariant tests across L1/L2** — the red tests + grouped-diff plumbing (`EvacuationDiffGroup` preserves per-command boundaries from the L2 response through `BlockResult`).
2. **Enforce double-entry accounting across the L1/L2 boundary** — conservation gate, deposit-gate equality, and the (since removed) runtime balance check.
3. **Sum Values with Value.combine instead of foldLeft** — mechanical sweep.
4. **Replace the runtime balance-identity check with builder property tests** — the design change above, plus `foldM`/`Acc` refactors in `StackEffectsBuilder`.
5. **Anchor the balance identity at the head boundaries** — init parse + crash recovery.

## Failure semantics

An ordinarily-invalid L2 tx is still rejected per request by the ledger (block continues). A *wrongly accepted* tx — the sugar-rush class — is caught deterministically on every peer at the next stack close, before any hard-ack is signed: the head fail-stops, no bogus commitment reaches L1, and funds remain evacuable under the last hard-confirmed committed map.


dragospe:
Pushed a third commit (`6956e04f`) tightening conservation from per-block to **per L2 command**.

The block-level check had a real hole: the L2 answers one command at a time, so diffs arrive with per-transaction boundaries — but `JointLedger` flattened them into `BlockResult.evacuationMapDiff`, and a block's aggregate delta can be zero while individual transactions err in compensating directions (a rounding error that over-credits buys and under-credits sells nets out under balanced flow). The new `compensating errors across two transactions are rejected` property demonstrates this red against the per-block check.

The per-command boundary is now preserved end to end (`EvacuationDiffGroup`: a `Transaction` group tagged with its producing request, or the block's `DepositDecisions` group), and `checkPartitionConservation` verifies each group against that command's own L1 boundary crossings, with a block-level aggregate identity as a shape backstop. `EvacuationMapNotConserved` now names the offending request.

Note: `BlockResult.evacuationMapDiff` changes shape, so this is a **persisted-format change to the `BlockResult` CF** (existing stores won't decode).

dragospe:
Restructured the branch into two commits (red/green) and completed the final tightening:

1. **`6498fcb9` — Add double-entry accounting invariant tests across L1/L2.** All invariant tests, red against this commit (16 failing checks): deposit-gate equality in both directions, per-command L2 value conservation (including compensating errors across two transactions in one block), withdrawal/absorption exactness, and the treasury balance identity — plus the diff-grouping plumbing needed to express per-command scenarios (shape only, no checks). Also includes a green probe through a real `SettlementTxSeq` build asserting the settlement builder maintains `value == map + equity + beacon` to the lovelace (it does).

2. **`9cdbf89b` — Enforce double-entry accounting across the L1/L2 boundary.** Turns everything green: deposit-gate conservation (`spawned == depositL2Value`), per-command conservation (`EvacuationMapNotConserved`, naming the offending request), and — new since the last push — the coverage bound tightened from `>=` to the exact balance identity `treasury.value == map + equity + beacon` (`TreasuryNotBalanced`): a deficit is insolvency, and a **surplus** is now also rejected — unaccounted value that the finalization's equity distribution would otherwise silently pocket, and that the init-tx parse (which takes the treasury output's value as-is) could smuggle in.

Full suite green; `-Werror`, scalafmt, scalafix clean. Reminder: `BlockResult.evacuationMapDiff` is grouped per command now — a persisted-format change to the `BlockResult` CF.

src/main/scala/hydrozoa/multisig/consensus/StackComposer.scala
```diff
@@ -1014,6 +1013,23 @@ object StackComposer {
                         lastBlockNum = unsignedStack.brief.lastBlockNum
                         treasury <- persistence.getOrFail(StoreKey.Treasury)
                         evacuationMap <- persistence.getOrFail(StoreKey.EvacuationMap(lastBlockNum))
+                        // The recovered pair is the balance-identity anchor re-entering the
+                        // system: the treasury and map snapshots live under two independent
+                        // store keys, so a divergent pair (partial write, replay bug, doctored
+                        // store) must be caught here, before the first stack close derives from
+                        // it. Like a missing key, an unbalanced pair is store corruption: fail
+                        // the boot.
+                        imbalance = treasury.value - evacuationMap.totalValue -
```
euonymos, line 1022: This goes against my intuition. Any doctored state in the database is outside our trust model - we trust the db. Then we should be sure we check it before writing.

dragospe: Hm, I think checking before reading _and_ writing probably is the most defensible?

From a formality standpoint, we'd want to establish the invariant at the boundary. If we re-architected a bit, we could even make this a type-level invariant via a smart constructor -- any "evacuation diff group" must satisfy the double-entry booking invariants, and we'd parse rather than validate..

PR #626, opened 2026-08-18
fix(l2-remote): implement restoreTo (send a Restore request to the remote ledger)

euonymos:
## The bug

A remote-L2-backed head **crash-loops on recovery**. `JointLedger.State.recover` co-anchors the L2 ledger via `l2Ledger.restoreTo(recoveredCommandNumber)`, but `RemoteL2Ledger.restoreTo` was a **no-op** (`EitherT.rightT(())`). So when the JointLedger's persisted command number lags the remote ledger's own durable position, nothing reconciles them: the JL re-issues an already-applied command, the remote's ordering gate expects `ledger.commandNumber + 1` and rejects it as `L2LedgerResponse.UnrecoverableError.OutOfOrder(sent, expected)`, `JointLedger.panicOnL2Response` raises, `TerminateActorSystem` fires, systemd restarts → same stale recovery → infinite crash loop. Observed live (`OutOfOrder(61844, 62028)`). The old no-op's own doc comment predicted exactly this.

## The change (client side only)

- New `Request.Restore(commandNumber)` request variant — carries only the command number (the number *is* the instruction; no command payload), following the existing request pattern.
- Its JSON codec in `RemoteL2LedgerCodecs`: a single-key tagged object `{"Restore": {"commandNumber": N}}`, matching the existing per-variant wire shape.
- A dedicated `RestoreResponse` ADT (`Restored(commandNumber)` / `RestoreFailed(commandNumber, reason)`) with its codec — single-key tagged objects `{"Restored": {"commandNumber": N}}` / `{"RestoreFailed": {"commandNumber": N, "reason": "..."}}`.
- `restoreTo` now sends `Request.Restore` through the existing `exchange` retry-through-transport path (mutex-serialised, command-number correlation preserved in a dedicated `sendRestoreRequest`), and interprets the response: `Restored` → `Right(())`, `RestoreFailed` → `Left(RestoreError.OtherError(reason))`. A genuine transport/protocol violation (undecodable frame, command-number echo mismatch) still fail-stops (a raise), exactly like the other three requests.

Surgical: only the remote-ledger client and its codecs change. `L2Ledger` / `L2LedgerResponse` / `RestoreError` and the EUTXO reference ledger are untouched.

### Design decision — response shape

I kept the restore response **out of** `L2LedgerResponse` (which models command *verdicts*: `Applied` / `Rejected` / `UnrecoverableError`). A restore is a boot-time reconstruction, not a numbered command verdict, so folding it into that ADT would have leaked a non-command concept into the verdict type and touched every other ledger. Instead I added a minimal `RestoreResponse` local to the remote protocol. It still carries `commandNumber` so the existing echo-correlation check holds (via a dedicated `sendRestoreRequest` mirroring `sendRequest`). A failure maps to `RestoreError.OtherError` — the closest existing `RestoreError` case (there is no remote-specific tip/desync info to justify `CommandNumberTooHigh`).

## **Requires SugarRush to implement the Restore request handler**

**This is a client-side change only. The remote server (SugarRush, a separate repo) must implement a handler for the new `Restore` request — rewind its command-number tip and committed state to `commandNumber` and answer `{"Restored": {"commandNumber": N}}` (or `{"RestoreFailed": ...}`). Until SugarRush supports it, this does NOT fix the live crash-loop end-to-end** — the request would go unanswered (retried through transport) or produce an undecodable/mismatched response (fail-stop).

## Compile

`sbt compile` — success (core + cardano-onchain).

### DIFF
--- a/src/main/scala/hydrozoa/multisig/ledger/remote/RemoteL2Ledger.scala
+++ b/src/main/scala/hydrozoa/multisig/ledger/remote/RemoteL2Ledger.scala
@@ -8,7 +8,7 @@ import cats.syntax.all.*
 import hydrozoa.config.head.network.CardanoNetwork
 import hydrozoa.lib.logging.ContraTracer
 import hydrozoa.multisig.ledger.l2.{ApplyDepositDecisionsResponse, ApplyTransactionResponse, L2CommandNumber, L2Ledger, L2LedgerCommand, L2LedgerResponse, RegisterDepositResponse, RestoreError}
-import hydrozoa.multisig.ledger.remote.RemoteL2Ledger.{Conn, Request}
+import hydrozoa.multisig.ledger.remote.RemoteL2Ledger.{Conn, Request, RestoreResponse}
 import hydrozoa.multisig.ledger.remote.RemoteL2LedgerEvent.*
 import io.circe.parser.*
 import io.circe.syntax.*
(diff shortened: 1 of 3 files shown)


euonymos:
Aligned the `restoreTo` wire shapes to SugarRush #140's canonical form (SugarRush's shape is canonical; this side changes to match):

- **Request** tag `Restore` → `RestoreTo`: `{"RestoreTo":{"commandNumber":N}}` (Scala variant stays `Request.Restore`; only the JSON tag changed).
- **Success**: `Restored(tip)` → `{"Restored":{"tip":N}}`.
- **Failure**: `RestoreFailed(requested, tip, reason)` → `{"RestoreFailed":{"requested":N,"tip":T,"reason":"…"}}`.
- Echo-correlation preserved via `RestoreResponse.commandNumber` (= `Restored.tip` / `RestoreFailed.requested`, both equal to the requested N).
- Failure mapping upgraded now that the shape carries both numbers: `RestoreFailed` with `requested > tip` → `RestoreError.CommandNumberTooHigh(requested, tip)`, otherwise → `RestoreError.OtherError(reason)`.

Added **golden codec tests** in `RemoteL2LedgerCodecsTest` pinning the exact JSON for all three frames (encode-equals-canonical + decode round-trip) — the cross-repo contract that would have caught this mismatch. All 12 tests in the suite pass.

PR #630, opened 2026-08-19
fix(ci): stop a failing CI run from hanging silently for hours

GeorgeFlerovsky:
A CI run that should have failed in minutes instead ran for **5 h 07 m** before being cancelled, of
which the integration suite was 4 h 57 m. It produced no error — it simply stopped making progress.
Five separate runs hung the same way.

This branch fixes the three defects behind that: the actor system could not report a failure and die,
the test tuning silently stopped applying, and nothing bounded the job.

### 1. An actor failure emergency-stopped the system instead of terminating it

**Symptom.** The actor system reported

```
Error(java.util.NoSuchElementException: None.get, kukku://…/user,
  class com.suprnation.actor.ActorSystem$UserGuardian,
  emergency stop: exception in failure handling for
  class hydrozoa.multisig.consensus.liaison.LaneOutbound$AppendOutOfOrder)
```

and then stopped. The JVM stayed **alive** with a dead actor system, every message going to
dead-letter — so nothing exited, nothing timed out, and anything awaiting the system waited forever.
Caught live on a 6-node head as well as in CI: a coil peer sat in this state while its head continued
without it. Under systemd this is worse than a crash, because the process never exits and is never
restarted.

**Mechanism**, confirmed by disassembling `cats-actors_3-2.1.0.jar` rather than inferred:

- `ActorSystem$UserGuardian.supervisorStrategy` is `TerminateActorSystem`, whose `decider` is
  `SupervisorStrategy.defaultDecider`.
- `defaultDecider` is defined for exactly four cases: `ActorInitializationException`,
  `ActorKilledException`, `DeathPactException` → `Stop`, and **`java.lang.Exception` → `Restart`**.
  Anything else is **undefined**.
- `SupervisionStrategy.handleFailure` calls `decider.applyOrElse(t, escalateDefault)`, and
  `escalateDefault` is total → `Escalate`. So a value extending `Throwable` but not `Exception`
  escalates *from the guardian*, which has no parent.
- `TerminateActorSystem.processFailure` opens with `Option.get` on the cause. On that path the option
  is empty, so the throw happens **inside failure handling** — the emergency stop above — and the
  orderly `TerminateActorSystem` never runs.

Most of our thrown values extended `Throwable` directly, which is invisible to that decider.

**Solution.** Make thrown values `Exception`-rooted, so the guardian's own directive applies and the
system terminates cleanly — which is the intended design: a failure takes the whole
`MultisigRegimeManager` subtree down, the process exits, systemd restarts it. Note this does *not*
change that design; `TerminateActorSystem.processFailure` terminates the system whatever directive it
is handed. It only decides whether it gets there orderly or via an emergency stop.

The refinement is worth doing on its own terms regardless: a meaningful exception type is the
convention the codebase already follows in most places, and a bare `Throwable` is what libraries
mishandle — cats-actors is one instance we happened to find, not the definition of the problem.

**Implementation** (`d146d76c`, `39445ac4`, `4066a928`, `49f0a090`), in four reviewable steps:

- **12 raise sites used `java.lang.Error` directly** — all saying "this actor reached its run loop
  without its connections". That is the JVM's category for `OutOfMemoryError` and friends, it is not
  an `Exception`, and unlike every other type here it is not matched by
  `scala.util.control.NonFatal`. Now `IllegalStateException`, which the codebase already uses for
  this exact shape (`CoilAckSequencer:64`, `RequestSequencer:49`, `BlockWeaver:89`). Messages
  unchanged.
- **14 types that are actually raised into `IO`** re-rooted at `RuntimeException` —
  `LaneOutbound.AppendOutOfOrder`, `SlowConsensusActor.CellError`,
  `FastConsensusActor.CollectingError` / `CompletionError`, `L2Ledger.RestoreError`, and the
  rule-based parse errors, each with its raise site cited in the commit.
- **The remainder**, used only as `Either` lefts, re-rooted for consistency.
- **`petri`'s three roots** (`Net.Error`, `Transition.Error`, `Place.MarkingError`). Nothing there can
  reach a supervisor — `core` does not depend on `petri` — but see the note above on why that is not
  the criterion.

After this the only bare `Throwable` left in any module, in any syntactic form, is
`SupervisionDeciderTest`'s deliberate one.

**Effect, measured.** `CoilRelayOrderingTest` provokes this exact raise. Before: `emergency stop` and
`None.get` once each. After: **zero of each**, and the failure appears as an orderly
`TerminateActorSystem` at the child. All three properties pass either way, so the test never depended
on the emergency stop — only on the raise.

### 2. The integration suite's tuning silently stopped applying

**Symptom.** `just integration-fast` died in about five seconds with a ScalaTest error — and
`.github/workflows/ci.yml` runs exactly that recipe.

**Mechanism.** The recipe was
`sbt "integration/testOnly * -- -s 10 -f ^(?!.*\(extended\))"`. sbt passes `--` arguments to **every**
test framework on the module. `integration` has ScalaCheck properties *and* ScalaTest suites, and
ScalaTest reads `-s` as "run this named suite" and throws. It had not bitten before only because
`build.sbt`'s `Tests.Exclude` happened to name every ScalaTest suite in that module — they are
excluded for needing Docker, so the arg protection was accidental. The first commit to trip it is
unmerged; adding any ScalaTest suite to `integration` breaks CI.

**Solution.** Scope the tuning to the framework it belongs to.

**Implementation** (`0145ec2c`). `build.sbt` gains
`Tests.Argument(TestFrameworks.ScalaCheck, "-s", "10", "-f", "^(?!.*\\(extended\\))")` and the args
leave the justfile. The unfiltered run stays available through the idiom the Yaci recipes already
use: `sbt "; set integration/Test/testOptions := Seq() ; integration/testOnly *"`.

Verified in **both** directions — the recipe no longer throws, *and* the tuning still applies
(`OK, passed 10 tests`, not 100, and zero `(extended)` properties). A fix that silently dropped the
tuning would look green locally and cost hours in CI.

### 3. Nothing bounded the job

**Symptom.** The 5 h 07 m run.

**Mechanism.** The `build` job carried no `timeout-minutes`, so it inherited GitHub's 6-hour default.

**Solution / implementation** (`0145ec2c`). `timeout-minutes: 90` on the `build` job. A full CI run is
about 45 minutes (fmt/lint 4 m 47 s + unit 6 m 05 s + integration 25 m 48 s + Yaci-docker 8 m 17 s),
so 90 leaves ample headroom while turning a hang into a bounded failure.

### Also here: a total supervision decider

`MultisigRegimeManagerBase`'s decider was a partial function that stopped at `case _: Exception`,
which reads as exhaustive and is not. It is now total, written with `PartialFunction.fromFunction`
(`8cf58d4e`), plus `SupervisionDeciderTest` (`78a472c5`) pinning what a supervised failure actually
does, by decider and by thrown class.

Two honest caveats. `PartialFunction.fromFunction` does **not** give a compiler warning for a missing
case — `Throwable` is not sealed — so this is a readability and belt-and-braces change, not a
compile-time guarantee. And it is not what fixes the hang: our decider escalates on every branch, so
it is never consulted for the value that kills the system. Defect 1 is the fix.

### Verification

- Core `Test/testOnly *`: **369 passed / 0 failed**, unchanged from this branch's base.
- **`petri/Test/testOnly *`: 139 / 0** — a module the default root-project invocation does not cover.
- `integration/Test/compile` clean; `cardanoOnchain/Test/testOnly *` clean.
- `fmtCheckAll` and `lintCheckAll` clean, run separately.
- The `AppendOutOfOrder` before/after measurement above.

### Notes for the reviewer

- **`RuleBasedActor.Error.BuildError` is an uninhabited `sealed trait`** — its six companion cases all
  extend `Unrecoverable` instead, and nothing pattern-matches on it. It is re-rooted here for
  consistency, but deleting it is probably better; left alone to keep this branch mechanical.
- **`CardanoBackend.Error(msg)`** now reads `extends RuntimeException(msg)`. Its `getMessage` override
  was already present and is untouched, so nothing observable changes — the exception's own
  `detailMessage` is simply no longer null.
- The catalogue behind this sweep is at `~/scratch/recovery-cycle/THROWABLES.md`, including the
  catch-site analysis: `NonFatal` already matched every bare-`Throwable` subclass, so defects in that
  category change no catch site; the `java.lang.Error` sites are the one category that becomes newly
  catchable, which is the direction we want.


integration/src/test/scala/hydrozoa/integration/stage1/Model.scala
```diff
@@ -809,7 +809,7 @@ object Model:
             yield ()
     }
 
-    enum Error extends Throwable:
+    enum Error extends RuntimeException:
         case UnexpectedState(msg: String)
```
dragospe: (non blocking): perhaps
```suggestion
    enum Error extends IllegalStateException:
case UnexpectedState(msg: String)
```

I suppose if we're getting more specific we should categorize accordingly?

PR #631, opened 2026-08-19
fix(recovery): six defects that stopped a restarted peer from rejoining

GeorgeFlerovsky:
A peer that was SIGKILLed and restarted did not reliably rejoin its head. This branch fixes the six
defects behind that, found by repeatedly crashing and restarting a live 6-node head (2 head + 4 coil
peers, real deposits, real settlements on preview) and by a deterministic crash-injection sweep.

Five of the six are the same defect class: **a linearity violation in recovery** — one fact reaching
a consumer by two paths, or a base that no path restores at all. The sixth is a timeout.

Stacked on `fix/ci-silent-hangs` (rebased onto its final tip); review that first. It also carries two
cherry-picked commits from `gh/test/recovery-crash-injection` (authorship preserved) — see *Notes for
the reviewer*.

---

### 1. A coil peer refused to boot: `store inconsistency (confirmed > acked)`

**Symptom.** After a crash, a coil peer would not start:
`Recovery refused: store inconsistency (confirmed > acked). softConfirmed=492933 fastBlockMark=492932`
— always exactly `+1`, and never on a head peer.

**Mechanism.** `JointLedger.handleBlock` forwarded the block brief to `FastConsensusActor` *before*
persisting the block bundle. A head peer survives that because its own soft-ack — authored after the
persist — also gates its consensus cell. A coil authors no soft-ack and already holds the head peers'
acks via its hub, so the brief alone saturates the cell and FCA writes `SoftConfirmation[n]` on its
own fiber, racing `persistCoilBlockBundle`. Crash inside that window and `softConfirmed` is one ahead
of `fastBlockMark`, which `ReplayActor.validateInvariants` refuses to boot on.

**Solution.** Persist before the hand-off, so the invariant holds by construction for both peer types
rather than only for head peers.

**Implementation** (`c888b6bd`). `handleBlock` restructured so the per-peer branch returns
`Option[SoftAck]` and only persists; the `fastConsensusActor ! brief` hand-off follows it; head-only
emissions happen last, inside `softAck.traverse_`.

### 2. `restoreTo` never completed, and retried forever

**Symptom.** A restarting peer looped indefinitely, each attempt restarting the remote ledger's work.

**Mechanism.** `RemoteL2Ledger.exchange` applied the ordinary 5 s command timeout to `Restore`, and
its failure path is drop-and-retry. SugarRush answers a restore by replaying its whole log — measured
at ~1.3 µs/event (probe added in the companion sugar-rush branch: 1.4 ms at 1k, 64 ms at 50k). Past a
few million events the bound expires, the retry re-sends, and the rebuild restarts from scratch.

**Solution.** Give `Restore` its own generous bound sized from that measurement, and make expiry
terminal rather than retried — a restore that times out is a fault to report, not to repeat.

**Implementation** (`792791ca`). Adds `restoreTimeout` (10 minutes), threads a per-request
`(timeout, retryOnTimeout)` through `exchange`, and adds a terminal branch emitting `RestoreTimedOut`.
Transport failures still retry, so a remote not yet listening at boot behaves as before.

### 3. BlockWeaver came back not knowing which block it was on

**Symptom.** Two failures, depending on how far the peer had got. Either
`TerminateActorSystem` — *"We are producing 2, but the confirmed block that we received is 5"* — or a
peer that reported **active**, admitted 24 000 requests until its mempool filled, and produced **zero**
blocks.

**Mechanism.** `BlockWeaver.State.start` hardcoded `nextBlockNumber = block 1`. A peer that had applied
everything it received gets nothing replayed (the ledger replay floor is `fastBlockMark + 1`), so it
opened at block 1 while the head was at block n — and armed itself as leader for block 2. Fixing only
the position exposed the second half: resuming at `fastBlockMark + 1` as leader means waiting for
`fastBlockMark`'s confirmation, but when `softConfirmed == fastBlockMark` — the ordinary case — both
replay floors sit above it and nothing ever re-derives it.

**Solution.** BlockWeaver reads its own base in its own `PreStart`, like every other actor with base
state, and applies the stored confirmation for the block it is resuming after.

**Implementation** (`903d06b5`). `BlockWeaver` takes `persistence`; its `PreStart` mirrors
`JointLedger.preStartLocal` — `initializeConnections`, then `Markers.recoverFastBlockMark`, then
`State.recover`. Two details are load-bearing: the stored confirmation is fed through
`state.react(config)` rather than constructing `Leader.AwaitingRequest` directly, because the
transition also arms the deadman `Wakeup` that forces a major block; and `react` can return `None`
(a final confirmation retires the weaver), so `PreStart` mirrors the become-loop's
`fold(context.self.stop)(become)`. Corrects three sentences in the persistence spec that claimed
`nextBlockNumber` comes from the replay tail.

### 4. The whole head's slow side stopped after any peer restart

**Symptom.** The most serious of the six, and silent. After a restart the fast side was flawless —
blocks flowed, majors happened, a fresh deposit absorbed — while **no stack hard-confirmed on any
peer**, including the peers that were never touched. Observed for 90 minutes and ~17 000 blocks. An
absorbed deposit sat at `SOFT_CONFIRMED` forever, so its settlement never reached L1. Nothing was
logged, because the slow-side tracers sit below the shipped log level.

**Mechanism.** `StackComposer` pairs a `BlockResult` with a `Block.SoftConfirmed` per block before
that block can enter a stack. `State.recover` rebuilt only the `BlockResult` half and left the
soft-confirmed halves "to replay". That tail cannot carry them: the block-spine replay floor is
`softConfirmed + 1`, while the blocks the next stack needs begin just after the last *closed* stack's
`lastBlockNum` — always below that floor, and far below once a round has stalled. So the follower
never paired those blocks, never covered the leader's brief, and waited in `tryCloseAsFollower`'s
benign "not yet covered" branch for events that had already happened and would never re-fire. Because
closing is gated on `previousStackHardConfirmed`, one peer's silence froze every peer.

**Solution.** Restore both halves from the store. Both are durable; only one was being read.

**Implementation** (`84714e08`). Adds `SoftConfirmationScan`, mirroring `BlockResultScan`, and folds
and pairs both halves in `State.recover` — the same thing `handleBlockResult` / `handleSoftConfirmed`
do live.

### 5. A replayed block brief reached FastConsensusActor twice

**Symptom.** A peer died on boot with
`CollectingError.UnexpectedBlock: Duplicate brief for block 44263`. Intermittent — it hit one coil
twice and its sibling never, and cleared on retry against the identical store.

**Mechanism.** `ReplayActor.route` sent every `Block` journal entry straight to FCA *and*, above the
ledger floor, to BlockWeaver — but `JointLedger.handleBlock` ends with an unconditional
`fastConsensusActor ! brief`. So one journal entry became two briefs, and `ConsensusCell.acceptBrief`
rejects a second brief for a block. The doubly-fed band is `[fastBlockMark+1, max(Block journal)]`,
non-empty only if the peer died after persisting an inbound brief and before writing that block's
`BlockResult` — hence the intermittency.

**Solution.** The two floors already describe two complementary bands; only one arm was gated by one.
Route each brief to exactly one consumer: below the floor this peer has already accepted the block,
so JointLedger will never re-handle it and only FCA — which tracks the *collective* mark and so lags —
needs it; at or above it, BlockWeaver weaves and JointLedger feeds FCA, exactly as on the live path,
where JointLedger is FCA's only source of briefs.

**Implementation** (`b935ecd8`). One `if/else` in `route`, plus a fail-loud assertion: above the floor
a brief can only be one this peer *follows*, because an own-led brief is written in the same atomic
batch as its `BlockResult` and cannot outlive it. That is what makes handing it to BlockWeaver safe —
a Leader state treats a `BlockBrief.Next` as `Unexpected` and panics — so the invariant is asserted
rather than assumed. `replay` gains an explicit `leadsFastBlock` predicate (`config.canLeadFast`).

### 6. Replayed hard-acks leaked into the orphan buffer

**Symptom.** No functional failure — an unbounded retention, growing with head lifetime and rebuilt on
every restart.

**Mechanism.** `handleRemoteHardAck` tells a late surplus ack from a genuinely-early orphan using
`State.lastConfirmed`. `State.initial` sets it to `None` and `SlowConsensusActor.preStart` read no base
state at all — it was the only consensus actor without one. With the guard inert, every replayed remote
ack for an already hard-confirmed stack was buffered as an orphan; orphans are cleared only at cell
creation, which never recurs for a stack that already confirmed. Since the hard-ack journals are
scanned from key 0, a restart retained every remote ack the head had ever produced.

**Solution.** Read the base — one marker — in the actor's own `PreStart`.

**Implementation** (`5fd2d2d6`). `Markers.recoverHardConfirmed` after `initializeConnections`, so the
base lands before the replay tail drains; `persistence` was already a constructor parameter.

---

### Test infrastructure

- **`DurabilityOrderOracle`** — a `Persistence` decorator that evaluates the boot gate
  (`confirmed <= acked`, both arms) after every durable write. It reads confirmed then acked
  *sequentially*: the marks are monotone, so an acked mark read second is never smaller than at the
  instant confirmed was read, which makes a reported violation sound. `Markers.derive` must not be used
  here — it reads with `parMapN` and manufactures false violations.
- **`SlowPersistence`** — N × `IO.cede` before each durable write. A same-peer fiber race does not
  reproduce under `TestControl`, whose scheduler runs the writing fiber straight through; yielding hands
  other fibers a turn exactly where a slow disk would, and costs no virtual time.
- **`DurabilityOrderTest`** — sweeps a topology ladder (1+0, 1+1, 1+2, 2+1, 2+2, 3+3) × yields (0, 2, 8).
  Against the pre-fix `JointLedger` every rung with ≥1 coil fails, at both non-zero yield settings,
  always `confirmed = acked + 1` and never on a head peer — matching the production logs exactly. The
  singleton passes. Post-fix all 18 pass.
- Unit regression tests for defects 4, 5 and 6, **each verified by a negative control** — reverting the
  fix must fail the test. This matters here: the pre-existing `RecoverSeamsTest` case for
  `StackComposer.recover` seeds only the `BlockResult` half, so it passed identically before and after
  defect 4's fix and could never have caught it.
- `RestartAfterBlocksTest` is committed **disabled**, with its reason: it livelocks on an unrelated
  known issue (`CardanoLiaison` builds a debug dump eagerly per tick, O(accumulated effects)).

### Verification

**Unit:** 374 passed / 0 failed. `fmtCheckAll` and `lintCheckAll` clean, run separately.

**In situ**, on a 6-node head with real deposits and settlements on preview:

- 20 minutes of load (12 req/s ledger-invalid + a rate-limited DEX bot at 100% valid orders),
  reaching 18 753 minor blocks, 3 major, 16 stacks, 5 absorbed deposits, 79 355 requests.
- SIGKILL of head-1 + coil-2 (JVM *and* ledger each, by process group), then of a head peer plus
  **both** its coils, then of the entire fleet — each followed by a restart against the existing
  stores. Every peer verified on the expected build from its boot banner.
- The sharpest evidence for defect 4: restarting **one** peer on the fixed binary closed the stuck
  stack and hard-confirmed it, and the cascade reached a peer still running the **old** binary after
  90 minutes of silence. The backlog then settled on L1 at block 4584561.
- Kill-all: stacks advanced past the pre-kill number and a deposit submitted *after* the restart
  absorbed, hard-confirmed, and settled on L1 at block 4584697.
- Re-run of kill-all against the final branch tip, with load running and a deposit already in flight:
  **zero failure signatures on all six peers**, stacks advanced past the pre-kill number, and the
  in-flight deposit absorbed and hard-confirmed after the restart, settling on L1 at block 4584858.
  The same coil had hit defect 5 on both earlier kill-alls and does not here.

### Notes for the reviewer

- **Stacked on `fix/ci-silent-hangs`** — merge or review that first.
- **Two cherry-picked commits** (`c3051d65`, `04d12096` after the rebase) are Ilia's, from the unmerged
  `gh/test/recovery-crash-injection`, taken with authorship preserved because this work needed the
  `CrashingPersistence` harness. If that branch lands first, rebase them away. The six production fixes
  are independent of them.
- **Deferred, deliberately:** a tight floor for the hard-ack replay arms. They are scanned from key 0
  because no persisted `StackNumber → HardAckNumber` correspondence exists to floor them
  (`ReplayCursors`, §10 Q9). With defect 6 fixed the replayed surplus is correctly dropped, so this is
  a waste-of-work item rather than a leak; a pointer comment sits on `route`'s `HardAck` arm.
- **Two ReplayActorTest assertions changed meaning** in defect 5 — they had pinned the double-feed
  (`hasBlock(c.fca, 5)` beside `hasBlock(c.bw, 5)`, and a seed comment reading "A block brief → FCA +
  BlockWeaver"). Worth a look: characterization, not intent, but it is a behaviour change to a test.
- The slow-side tracers log under their own top-level logger names, so a stalled stack round is
  completely silent at the shipped log level. That is why defect 4 survived so long, and is worth
  considering separately from this PR.


integration/src/test/scala/hydrozoa/integration/harness/MultiPeerHeadHarness.scala
```diff
@@ -717,19 +731,39 @@ object MultiPeerHeadHarness:
...
+            coilTicks <- Resource.make(
+              Ref[IO].of(Map.empty[CoilPeerNumber, IO[Unit]]).flatTap { ref =>
+                  coilConnections.toList.traverse_ { case (coilNum, conns) =>
+                      Ticks
+                          .tickLoop(coilPollingPeriodOf(coilNum), conns)
+                          .start
+                          .flatMap(fib => ref.update(_.updated(coilNum, fib.cancel)))
+                  }
+              }
+            )(_.get.flatMap(_.values.toList.traverse_(identity)))
```
dragospe, line 766: There's an issue with these, pushing a fix.

src/main/scala/hydrozoa/multisig/HeadMultisigRegimeManager.scala
```diff
@@ -234,7 +234,8 @@ trait HeadMultisigRegimeManager(
               peers = config.headPeerIds.map(_.peerNum).toList,
               hubs = config.hubHeadPeerNumbers,
               coils = hubbedCoilPeers,
-              treasuryAddress = config.initializationTx.treasuryProduced.address
+              treasuryAddress = config.initializationTx.treasuryProduced.address,
+              leadsFastBlock = config.canLeadFast
```
dragospe, line 238: Just a nit, if all of these can be derived from `config`, they should be constructed inside the method via the `using config`.

integration/src/test/scala/hydrozoa/integration/harness/CoilCrashRestartTest.scala
```diff
@@ -0,0 +1,129 @@
...
+import org.scalacheck.rng.Seed
+import org.scalatest.funsuite.AnyFunSuite
+import scala.concurrent.duration.*
+
+/** The coil-side counterpart to [[CrashRestartTest]]: crash a coil peer at a chosen durable write,
+  * restart it against its own store, and assert the head goes on hard-confirming.
+  *
+  * Coil recovery had no end-to-end exercise at all — the harness could not even restart a coil
+  * (`restartCoilPeer` and the `wrapPersistence` hook on `buildCoil` are new), which is part of why
+  * the fleet's coil peers were the ones that would not come back.
```
dragospe: Your claude writes a lot of comments, I'd delete this one. Historical commentary usually belongs in PR/commit notes, and doesn't age well as code comments.


```suggestion
```

src/main/scala/hydrozoa/multisig/consensus/BlockWeaver.scala
```diff
@@ -30,9 +31,19 @@ final case class BlockWeaver(
         BlockWeaver.ConnectionsPartial,
     tracer: ContraTracer[IO, BlockWeaverEvent],
     metrics: PeerMetrics,
+    /** Read-only here: the weaver authors no journal. It still has base state to recover (§5.2) —
+      * which block to resume on, and whether the one before it was already confirmed — and, like
+      * every other actor with base state, it reads its own in [[receive]]'s `PreStart`.
+      */
+    persistence: Persistence[IO],
```
dragospe: We should probably expose a `ReadPersistence` and `WritePersistence` trait if its important that this is read-only.

src/main/scala/hydrozoa/multisig/consensus/BlockWeaver.scala
```diff
@@ -196,10 +218,80 @@ object BlockWeaver {
...
+          * it does on the live path. When `softConfirmed < fastBlockMark` the predecessor genuinely
+          * is not confirmed yet and the `FastConsensusActor` re-derives it as the replayed tail
+          * re-saturates its cells, so this is a no-op — as it is on a follower, which waits on
+          * briefs rather than confirmations, and on a cold store.
+          */
+        def recover(
+            config: Config,
+            connections: Connections,
+            tracer: ContraTracer[IO, BlockWeaverEvent],
+            persistence: Persistence[IO],
```
dragospe, line 231: IMO these should all be in a `using` clause. They're ambient, add clutter, and get passed through unmodified/inspected.

src/main/scala/hydrozoa/multisig/consensus/BlockWeaver.scala
```diff
@@ -196,10 +218,80 @@ object BlockWeaver {
...
+          *
+          * Whether the block before `fastBlockMark + 1` is already confirmed is the other half of
+          * the base state; [[BlockWeaver.resumePreviousConfirmation]] applies it to the state this
+          * returns.
+          */
         def start(
             config: Config,
             connections: Connections,
-            tracer: ContraTracer[IO, BlockWeaverEvent]
+            tracer: ContraTracer[IO, BlockWeaverEvent],
```
dragospe, line 263: likewise here

src/main/scala/hydrozoa/multisig/consensus/BlockWeaver.scala
dragospe: Ticket: https://linear.app/gummiworm-labs/issue/GUM-281/refactor-blockweaver-to-use-using-clauses

src/main/scala/hydrozoa/multisig/consensus/ReplayActor.scala
```diff
@@ -73,7 +76,8 @@ object ReplayActor:
         peers: List[HeadPeerNumber],
         hubs: List[HeadPeerNumber],
         coils: List[CoilPeerNumber],
-        treasuryAddress: ShelleyAddress
+        treasuryAddress: ShelleyAddress,
+        leadsFastBlock: BlockNumber => Boolean
     )(using CardanoNetwork.Section): IO[Unit] =
```
dragospe, line 79: Likewise, a number of these can be condensed into a `using`

dragospe: https://linear.app/gummiworm-labs/issue/GUM-282/refactor-replayactor-to-use-using-clauses

src/main/scala/hydrozoa/multisig/CoilMultisigRegimeManager.scala
```diff
@@ -125,7 +125,8 @@ trait CoilMultisigRegimeManager(
               // A coil peer is never a hub, so it re-feeds no coil-ack gap (its Targets carries no
               // CoilAckSequencer either — the gap step is a no-op).
               coils = Nil,
-              treasuryAddress = config.initializationTx.treasuryProduced.address
+              treasuryAddress = config.initializationTx.treasuryProduced.address,
+              leadsFastBlock = config.canLeadFast
             )(using config)
```
dragospe, line 130: same here

dragospe (commented):
Note: I was only looking at the app code rather than the tests.

My biggest observation is that we're growing into a noisier style. A lot of static config and immutable-once-set runtime-state (tracers, persistence, etc) is being threaded through manually in a lot of places, and this is harming readability.

I made a few tickets, but in the future I think we should clean this up and rely more on `(using config, actorState)` in more places.

PR #633, opened 2026-08-19
Screen deposits and transactions via the remote ledger's screening endpoint

dragospe:
## What

The client side of remote screening, for **both deposits and transactions**: with `remoteScreenerUri` configured, a remote-ledger node screens each request against the SugarRush screener (**GummiwormLabs/sugar-rush-ledger#149**) *before* assigning it a RequestId — a malformed request is rejected up front with the screener's reason instead of riding the full consensus round trip.

Closes the `any-remote` screening gap documented in `docs/spec/l2-isomorphism.md` (Limitations): the passthrough `RemoteL2Screener` object becomes a real HTTP client for both `screenDeposit` and `screenTx`.

## Semantics

- **Rejection is a verdict**: `{ok: false, reason}` → `L2ScreenError(reason)`, refused pre-RequestId, surfaced to the user.
- **Transport failure fails open**: refused connection, timeout, non-2xx, undecodable body → the request proceeds *unscreened*, exactly as if no screener were configured (warn-level trace). Screening is advisory — the remote ledger re-runs the same checks authoritatively at submission — so a sidecar outage must not block the request path the way the mutation transport's retry-forever policy would. The doc comment spells out this asymmetry.
- **Deposit and transaction share one path**: `screenDeposit` and `screenTx` both post through a single private `screen` helper — same verdict mapping, same fail-open, differing only in endpoint, request body, and the rejection event traced (`DepositRejected` / `TxRejected`).
- The ember client is dedicated (screening never shares the mutation transport) with a **5s timeout** — a hung screener costs a request seconds, not ember's 45s default.

## What each endpoint screens

- **`POST /screen/deposit`** — the ledger checks the deposit's `l2Payload` is well-formed and its declared value matches the outputs it will spawn (after Hydrozoa's own L1 pin + accept-by screening).
- **`POST /screen/tx`** — the ledger decrypts the payload, verifies its COSE signature over the exact body bytes, then runs the parameter-only stateless tx checks: admin authorization, market fees ∈ [0, 1), nonzero order price, no ephemeral-asset withdrawal, and delegation changes account-key signed. (These land on the SugarRush side in #149.)

## Wire contract

`POST /screen/deposit` carries `ScreenDeposit` — the `RegisterDeposit` fields minus the consensus-assigned `requestId`/`blockNumber`/`blockCreationStartTime`, in the same SugarRush encodings the ws mutation path uses. `RegisterDeposit`'s encoder now *delegates* to the `ScreenDeposit` encoder and deep-merges the consensus fields, pinning "a RegisterDeposit is a ScreenDeposit plus consensus fields" structurally so the two bodies cannot drift.

`POST /screen/tx` carries just the encrypted `l2Payload` (`screenTxBody`) — a transaction self-authenticates through its own witnesses, so it carries none of the deposit's reference fields.

Both bodies are pinned by golden tests against the exact JSON the SugarRush side pins in its own goldens (`screener/src/screen.rs::screen_deposit_request_golden` / `screen_tx_request_golden`, `types/.../register_deposit.rs::register_deposit_golden`) — a cross-repo contract test pair.

## Config

`remoteScreenerUri: Option[String]` in the node's private config, next to `remoteLedgerUri`. Absent → passthrough (existing `private.json` files decode unchanged). New `RemoteL2Screener` tracer route, logback configs updated in sync.

## Testing

- `RemoteL2ScreenerTest`: deposit + tx wire goldens, the ok/reject/reasonless-reject verdict mapping, non-2xx and transport-failure fail-open (both request kinds), and that a `screenTx` POST hits `/screen/tx`.
- New `RegisterDeposit` golden in `RemoteL2LedgerCodecsTest` (the command's full wire form had no coverage before the encoder rework).
- `just build-werror`, `fmt-check`, `lint-check` clean; remote + config suites green.

## Included wallet fix

This branch also carries the two commits from **#650** (retire `WalletModule.BloxBean` for vanilla Scalus keys; splice peer 0's real signing key into `MainSmokeTest`'s config so the node can sign) — cherry-picked here to keep CI green. They land independently via #650 and fall out of the diff once this branch merges `main`.

## Deployment note

To turn this on for the demo head, the sugar-rush `head/demo` peer-private template needs `"remoteScreenerUri": "http://127.0.0.1:3003"` (the screener joins the head-0 pod in GummiwormLabs/sugar-rush-ledger#149) and the compose file needs an image built from this branch.


src/main/scala/hydrozoa/multisig/ledger/l2/L2LedgerCommand.scala
```diff
@@ -76,32 +76,39 @@ object L2LedgerCommand {
...
-            import hydrozoa.multisig.ledger.remote.RemoteL2LedgerCodecs.{given_Decoder_Value, given_Decoder_Coin}
-            import RequestId.i64.given // L2-ledger / SugarRush wire form (i64), not the default object
-            io.circe.generic.semiauto.deriveDecoder
-        }
+        // Deliberately encode-only: Hydrozoa only ever *sends* a RegisterDeposit (the remote
+        // ledger decodes it with its own types). A decoder could not invert the encoder anyway —
+        // `Destination` is encoded as a lossy `{address, datum}` object while its circe decoder
+        // reads a CBOR-hex string. The one path that needs an exact RegisterDeposit round-trip,
+        // the EUTXO ledger's durable command log, re-derives symmetric store-local codecs
+        // (`eutxol2.store.L2StoreCodecs`) instead of using this wire form.
```
dragospe: Delete this comment

src/main/scala/hydrozoa/multisig/ledger/l2/L2LedgerCommand.scala
```diff
@@ -76,32 +76,39 @@ object L2LedgerCommand {
...
+                      l2Payload = r.l2Payload
+                    )
+                    .asJson
+                    .deepMerge(
+                      io.circe.Json.obj(
+                        "requestId" -> r.requestId.asJson,
+                        "blockNumber" -> r.blockNumber.asJson,
+                        "blockCreationStartTime" -> r.blockCreationStartTime.asJson
+                      )
+                    )
```
dragospe: Might have performance impact due to double serialization? Perhaps do it in one pass.

PR #643, opened 2026-08-20
Let an opening L2 output carry an explicit evacuation key

euonymos:
Recovered from an uncommitted working tree on the demo box — **authored by @Pi-Lanningham**, committed here so it stops living one `git checkout` from oblivion. It is what the running demo head was actually built with.

## Why

`build-head-config` keys the opening evacuation map by a synthetic eutxo reference derived from the seed utxo. A remote L2 ledger keys its own map by **account hash** — Sugar Rush uses a 28-byte key hash zero-padded to 32. With no way to express that, an `any-remote` head commits an `initialEvacuationMap` its ledger never agrees with: different key space, silently.

This is not theoretical. It is why the demo head needed a hand-modified binary to build a bootable config, and — with cardano-hydrozoa#639 — a head built without it now refuses to start.

`L2Output` gains an optional `evacuationKey`. Entries that set it keep it; entries that don't keep the synthetic convention. Obligation construction, including the min-ada check, is untouched.

## Verified

On a branch carrying both this and #639, the competition bootstrap directory produces:

```
e04159fa…f84400000000 -> a200581d60e04159fa…f844011a00989680
digest: 66beaaaaa0c2c624db9b83eb21bde446580944d825bb6ac2ea5f76c5d4633d29
```

byte-identical to what `sugar-rush-ledger-server print-initial-evacuation-map` reports for the matching ledger config. Two independent implementations, same bytes.

## Review notes — this is a stopgap, and the code says so

The patch's own comment reads *"Remove once build-head-config grows a real any-remote mode."* Agreed, and worth recording why before it calcifies:

- **It duplicates derivable data.** For Sugar Rush the key and the address are the same account hash — note above that the map key's first 28 bytes and the output's address bytes coincide. Operators supply both halves of one fact with nothing checking they agree.
- **`L2Output` is the wrong home.** The type means "an output in the EUTXO L2 ledger". A remote ledger has accounts, not outputs.
- **It cannot express a destination datum.** Sugar Rush's `Destination` is `{address, datum: Option<Datum>}`, and `toTransactionOutput` hardcodes `datumOption = None`. The whitepaper's ledger-state page says that datum is what "enables treasury-owned sugar rush accounts, with trading authority temporarily delegated". Route such an account through `L2Output` and you get different CBOR, a different digest, and a head that won't boot — or, pre-#639, a committed map paying somewhere else. Today's fee account has no datum, so we're fine; it's a landmine, not a wall.

The durable fix is an `initial-evacuation-map.json` bootstrap input: the map verbatim in the head config's own encoding, exactly what the ledger already prints. No re-derivation, no redundant key, datums preserved because the CBOR is never decoded and rebuilt. That would also retire the `sed` workaround for `l2Ledger` that sugar-rush's `DEPLOYMENT.md` §3 currently documents.

Landing this now because the competition launch needs a bootable head config today.

## Tests

Four cases on the decoder: an explicit key is read, absence leaves the synthetic convention, a malformed key is rejected rather than dropped (circe ignores unknown fields, so a typo would otherwise pass unnoticed), and address/value still decode alongside a key.

`just fmt`, `just lint`, and `just build-werror` all clean. The original patch left `Utxos` unused and imports unsorted; both fixed.

### DIFF
--- a/src/main/scala/hydrozoa/bootstrap/Bootstrap.scala
+++ b/src/main/scala/hydrozoa/bootstrap/Bootstrap.scala
@@ -30,10 +30,15 @@ import hydrozoa.multisig.backend.cardano.{CardanoBackend, CardanoBackendBlockfro
 import hydrozoa.multisig.consensus.peer.HeadPeerNumber
 import hydrozoa.multisig.consensus.peer.HeadPeerNumber.given
 import hydrozoa.multisig.ledger.block.{Block, BlockBrief, BlockEffects, BlockHeader}
-import hydrozoa.multisig.ledger.eutxol2.toEvacuationMap
+import hydrozoa.multisig.ledger.eutxol2.toEvacuationKey
 import hydrozoa.multisig.ledger.eutxol2.tx.L2Genesis
+import hydrozoa.multisig.ledger.joint.EvacuationMapInstances.given
+import hydrozoa.multisig.ledger.joint.given
+import hydrozoa.multisig.ledger.joint.obligation.Payout
+import hydrozoa.multisig.ledger.joint.{EvacuationKey, EvacuationMap}
 import hydrozoa.multisig.ledger.l1.tx.RawTx
 import hydrozoa.multisig.ledger.l1.txseq.InitializationTxSeq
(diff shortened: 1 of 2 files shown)


euonymos:
Superseded by #666.

PR #647, opened 2026-08-20
Stop sbt silently dropping test failures on forked runs

GeorgeFlerovsky:
On `main`, a falsified ScalaCheck property can be reported as passed with exit code 0. It is not rare: in the core suite, **95 of 141 property results never reach sbt's tally** — exactly one survives per suite.

```
                 ScalaTest   reported total   "+" printed   "!" printed
main                   351          397           141             0
this branch            351          492           141             0
```

Core has 46 `Properties` suites, and 351 + 46 = 397: main counts one property per suite and discards the rest. Whether the survivor is a pass or a failure depends on arrival order.

Both runs print all 141 property results and no falsifications, so nothing is failing and being hidden today — this restores the tally's integrity rather than uncovering broken tests.

## Why

Two things compose.

sbt's forked worker reports a task's events keyed on `taskDef.fullyQualifiedName()`, and the build side stores them with `results += info.group -> suiteResult` into a `mutable.Map` — a second batch for the same name **replaces** the first instead of merging.

ScalaCheck normally avoids producing several batches per suite: since 2018 it takes a single-task path when it detects a forked run, detected by looking for `ForkMain` in the fingerprint's class name. sbt 2 renamed that class to `sbt.internal.worker1.ForkTestMain$SubclassFingerscan`, so the check stops matching, ScalaCheck falls back to one task per property — all sharing the suite's fully-qualified name — and the events overwrite each other.

Isolated to the sbt version, one variable changed: sbt 1.11.3 reports `Failed: Total 4, Failed 1` and exits 1; sbt 2.0.1 reports `Passed: Total 1` and exits 0. Both filed upstream: typelevel/scalacheck#1195 and sbt/sbt#9642.

## What this does

**`ScalaCheckFrameworkFixed`** delegates every call to `ScalaCheckFramework` and changes one thing: where a suite's events are delivered from. `rootTask.execute` only *enumerates* the per-property tasks, so the wrapper runs them itself against the caller's handler and they leave as a single batch. They still run concurrently, on one shared bounded pool, so no intra-suite parallelism is lost — the probe below shows no slowdown.

It needs no access to ScalaCheck internals, so no fork or vendored copy.

It lives in the root-level `test` package, alongside `Generators` / `TestPeers` / `SeedPhrase`, and **must not move to `hydrozoa.test`**: that shadows the top-level `test` package for every file under `hydrozoa.*`, so `import test.Generators...` stops resolving. Incremental builds stay green and only a clean compile shows it — I made exactly that mistake writing this.

**`checkScalaCheckFramework`** guards the registration. That is the fragile part: forgetting it would restore the old behaviour silently, and the symptom is a green build — the same signature as the bug. The check inspects **every** project, so it also covers a project that never opted in, which a per-project check would miss by construction. `just precommit` runs it, so CI does.

**#646 already fixed the same overwrite on the ScalaTest side** and is now in `main`, so this PR builds on it and touches none of that ground.

One thing that PR's description understated, worth recording since it is the same root cause: the duplicate registration did not merely waste time, it **hid failures**. Both runs of a suite report under one suite name, so the `results += group -> suiteResult` overwrite below kept just one — observed here as 342 test names each printed twice, one pass reporting `failed 1`, and the aggregate reporting `Failed 0` with exit code 0 (the `Serve.runNode … ServerStarted` case). #646 closed a real correctness hole, not only a wall-clock one.

`integration`'s ScalaCheck `testOptions` are re-scoped to the new framework. `testOptions` are matched *by framework*, so leaving them pointed at `TestFrameworks.ScalaCheck` would have silently dropped the tuning — 100 cases instead of 10, and `(extended)` properties back in the fast run.

## Verification

On this commit, against `main` @ `26084076`:

- Core suite green — **492 results reported where `main` reports 397**, 141 `+` lines, zero `!`, and a single `Tests:` summary.
- `checkScalaCheckFramework` reports no offenders.
- `fmtCheckAll` and `lintCheckAll` clean, run separately.

On the same diff before the final rebase (the rebase replayed it unchanged onto a newer base):

```
probe suite, one fast failing property + three slow passing ones
  with the wrapper        Failed: Total 4, Failed 1   exit 1   10.2s
  main's build.sbt        Passed: Total 1, Failed 0   exit 0   12.3s
```

- Negative control on the guard: with the registration removed, `checkScalaCheckFramework` fails and names all five projects carrying ScalaCheck on their classpath.
- `integration`'s tuning still applies — a probe property reports `OK, passed 10 tests` and its `(extended)` sibling is filtered out.
- Clean multi-root compile with zero compiler warnings.

**This changes no test's outcome.** Zero `!` lines on both `main` and this branch, so every one of the 95 results `main` discards is a *pass* — the exposure is real but unrealised, and no test needed fixing.

**Not run here:** the full integration suite (~25 min), left to CI. That is where the remaining `Properties` suites and the stateful stage suites live, so it is the first place a previously-hidden failure would surface.

## Removing this

All three pieces — the wrapper, its registration, and the check — come out once `build.sbt` resolves an sbt or ScalaCheck version carrying either upstream fix. The check's failure message says so, with both issue numbers.

## Reviewer notes

- `checkScalaCheckFramework` is `Def.uncached`: it reads `.all(...)` across projects, which sbt 2 cannot hash for task caching. It is wired into `just precommit` rather than `Test / test` deliberately — `Test / test` is an `InputTask` in sbt 2, and redefining `Test / executeTests` instead would have forced test caching off.
- The settings sit next to their own definitions rather than beside the (now removed) ScalaTest registration, so this PR's `build.sbt` hunks stayed disjoint from #646's throughout.


GeorgeFlerovsky:
### Follow-up commits: the CI failure this PR surfaced

CI on the first commit reported `Failed: Total 492, Failed 1`. The 492 is this PR working; the
failure was `MainSmokeTest`, whose node dies on **every** run. Root cause and full analysis in #649 —
@dragospe filed it independently the same day, and these commits are his option 1.

The key cannot simply be spliced back into the config: `TestPeers` wallets are BIP32 extended keys
with no 32-byte form. So the fix gives that one caller a key that survives the JSON round trip.

- `WalletModule.exportSigningKey: Option[ScalusSigningKey]` — `Some` for `Scalus`, `None` for BloxBean.
- `dummyPeerWalletEncoder` → renamed **`peerWalletEncoder`**: writes the real key when there is one.
- `TestPeers.KeyScheme.{Bip32, Ed25519}`, default `Bip32`, so no existing suite's keys move. The
  vkey/wallet caches now key on scheme too — without that, one `Ed25519` instance poisons every
  BloxBean suite sharing a mnemonic, Yaci ones included.
- Only `MainSmokeTest` opts in. `GenerateSampleConfig`'s CLI stays on BIP32: its `defaultSpec` uses
  `SeedPhrase.Yaci`, whose derivation lands on the devnet's pre-funded addresses.
- The assertion now requires the node to still be alive after the bind, and treats termination —
  exit code 0 included — as failure. A green result was the whole problem.

Two production changes, because they are why this read as a flake rather than a crash.
`Serve.runHeadNode`/`runCoilNode` returned `ExitCode.Success` after `waitForTermination`, so a
guardian escalation exited 0 and no supervisor would restart it; they now return `ExitCode.Error`.
That is sound because nothing in the node calls `system.terminate`. The cause itself cannot be
logged: cats-actors 2.1.0 records it, then its own `Resource.onFinalize(terminate(None))` clears it
before the finalizer that would re-raise it.

One oddity to expect in the diff: the test ends in `fiber.cancel.start.void`, because cancelling a
live node never returns and `Fiber.cancel` is uncancelable, so no timeout bounds it.

### Verification

From a clean build of every root:

- Core suite `Passed: Total 492, Failed 0` / `succeeded 351, failed 0` on three runs; `main`'s CI run
  of the same suite reported `succeeded 350, failed 1`. A fourth run turned up one property failure —
  see below, and it is not this branch.
- `MainSmokeTest` logs no `BadSignature`. Negative control: flipping back to `KeyScheme.Bip32`
  fails it, where before it passed.
- Every `[EventBus] => Error` in a full local run is a deliberate provocation; the `hydrozoa-demo`
  one is gone.
- Zero `[warn] -- [E` across both `Test/compile` roots, so green under `-Werror`. `fmtCheckAll` and
  `lintCheckAll` clean.
- Integration red on one unrelated result: `RBR Seed Abstraction.ws`, a 7-minute `TimeoutException`.
  This PR neither unmasks it (single-property suite, so `main` reports it identically) nor slows it —
  isolated it takes 218 s here against 263 s on `main`, both under a 420 s budget. The margin is the
  problem.

### The reporting fix caught a real failure

That fourth run came back `Failed: Total 492, Failed 1` — the same tally as the CI failure at the
top, but a different thing failing: a *property* this time, not a ScalaTest case. It is worth
recording because every piece of evidence in the description is a dropped *pass*:

```
! TransientTokens conservation.removing one declared token fails the main projection:
    Falsified after 3 passed tests.
  expected an main-projection conservation failure, got:
    Left(TransactionException$InvalidSignaturesInWitnessesException: ...)
```

The property mutates a scenario's declarations and asserts on one specific rejection message, but
mutating the tx body lets the signature check reject it first — correct rejection, wrong reason. It
is unrelated to this branch (the suite's config comes from a fixed `Seed(0L)`; only its own generator
is unseeded) and rare: one failure, then six clean isolated repeats.

**`TransientTokensConservationTest` declares three properties**, so on `main` all three tasks share
one fully-qualified name and roughly two times in three this failure is discarded and the suite
reports green. So the description's "every one of the 95 results `main` discards is a *pass*" holds for the runs it
measured, and this is the counter-example: the mechanism drops real failures too, not just passes.

GeorgeFlerovsky:
### Follow-up commits: the CI failure this PR surfaced, and whose fix landed

CI on the first commit reported `Failed: Total 492, Failed 1`. The 492 is this PR working; the
failure was `MainSmokeTest`, whose node dies on **every** run — root cause and full analysis in #649,
which @dragospe filed independently the same day.

I first fixed it here with an opt-in key scheme on `TestPeers`. **#650 does it better** — retiring
`WalletModule.BloxBean` outright, 3 files and a net deletion against 12 files and an abstraction, and
one fewer test-only outlier in `src/main`. So that branch is merged in here and my version is
reverted: `TestPeers.KeyScheme`, `WalletModule.exportSigningKey`, `PeerWallet.exportSigningKey` and
the encoder rename are all gone. The encoder still withholds the signing key; `MainSmokeTest` splices
it back, exactly as `GenerateKeyPair` does for real configs.

**This PR is stacked on #650**, so the diff above shows only what this branch adds on top of his.
#650 merges first; GitHub retargets this one to `main` when it does.

What this branch still contributes beyond the reporting fix:

- `Serve.runHeadNode`/`runCoilNode` returned `ExitCode.Success` after `waitForTermination`, so a
  guardian escalation exited 0 and no supervisor would restart the node. They now return
  `ExitCode.Error`. Sound because nothing in the node calls `system.terminate`. The cause itself
  cannot be logged: cats-actors 2.1.0 records it, then its own
  `Resource.onFinalize(terminate(None))` clears it before the finalizer that would re-raise it.
- `MainSmokeTest`'s assertion requires the node to **outlive** the bind, rather than racing the
  milestone against `runNode` returning — a green result was the whole problem. It pairs with a
  fire-and-forget `fiber.cancel.start.void`: awaiting the cancel hangs once the node has been live a
  few seconds, and `Fiber.cancel` is uncancelable so no timeout bounds it. Keep the settle window and
  you must not await the cancel.

### Verification

From a clean build of every root:

- Core suite `Passed: Total 492, Failed 0` / `succeeded 351, failed 0`, unchanged across the merge;
  `main`'s CI run of the same suite reported `succeeded 350, failed 1`. One earlier run turned up a
  property failure — see below, and it is not this branch.
- `MainSmokeTest` logs no `BadSignature`. **Negative control on the merged code:** drop #650's splice
  and the test *fails* — `BadSignature` returns, the node dies, and the strengthened assertion catches
  it (`reached ServerStarted but the node then terminated`). That is the two halves working together:
  the splice keeps the node alive, the assertion is what notices if it ever stops being.
- Every `[EventBus] => Error` in a full local run is a deliberate provocation; the `hydrozoa-demo`
  one is gone.
- Zero `[warn] -- [E` across both `Test/compile` roots, so green under `-Werror`. `fmtCheckAll` and
  `lintCheckAll` clean.
- Integration red on one unrelated result: `RBR Seed Abstraction.ws`, a 7-minute `TimeoutException`.
  This PR neither unmasks it (single-property suite, so `main` reports it identically) nor slows it —
  isolated it takes 218 s here against 263 s on `main`, both under a 420 s budget. The margin is the
  problem.

### The reporting fix caught a real failure

That fourth run came back `Failed: Total 492, Failed 1` — the same tally as the CI failure at the
top, but a different thing failing: a *property* this time, not a ScalaTest case. It is worth
recording because every piece of evidence in the description is a dropped *pass*:

```
! TransientTokens conservation.removing one declared token fails the main projection:
    Falsified after 3 passed tests.
  expected an main-projection conservation failure, got:
    Left(TransactionException$InvalidSignaturesInWitnessesException: ...)
```

The property mutates a scenario's declarations and asserts on one specific rejection message, but
mutating the tx body lets the signature check reject it first — correct rejection, wrong reason. It
is unrelated to this branch (the suite's config comes from a fixed `Seed(0L)`; only its own generator
is unseeded) and rare: one failure, then six clean isolated repeats.

**`TransientTokensConservationTest` declares three properties**, so on `main` all three tasks share
one fully-qualified name and roughly two times in three this failure is discarded and the suite
reports green. So the description's "every one of the 95 results `main` discards is a *pass*" holds for the runs it
measured, and this is the counter-example: the mechanism drops real failures too, not just passes.

PR #650, opened 2026-08-20
Retire WalletModule.BloxBean; fix MainSmokeTest flakiness (runnable test config)

### LINKED ISSUE #649: MainSmokeTest flaky: node boots with dummy (zero) signing key from serialized TestPeers config → deterministic stack-0 hard-ack BadSignature racing ServerStarted
## Summary

`hydrozoa.app.MainSmokeTest` ("Serve.runNode reaches ServerStarted for a generated single-peer config") is flaky in CI. It is **not** a timing-only flake: the node **deterministically** fails to verify its own stack‑0 hard‑ack signature on every run because the generated on‑disk config carries a **dummy all‑zero signing key**. The only thing that is racy is whether the HTTP server binds (`ServerStarted`) before the resulting consensus‑break terminates the actor system.

Reproduces on `main` (measured ~1 in 3–4 runs fail locally). It is what turns PRs red — e.g. #633 — without being related to the PR's own diff.

## Symptom

```
- Serve.runNode reaches ServerStarted for a generated single-peer config *** FAILED ***
  java.lang.AssertionError: Serve.runNode terminated before reaching ServerStarted: Succeeded(IO(ExitCode(0)))
  at MainSmokeTest.scala:107
```

Underlying, every run logs:

```
SlowConsensusActor$CellError$BadSignature: Hard-ack signature verification failed for tx "…"
  → com.suprnation.actor.TerminateActorSystem
```

## Root cause

1. `MainSmokeTest` generates a single-peer `MultiNodeConfig` via `TestPeers`, writes `head-config.json` + `peer-0/private.json` to a temp dir, and boots the **real** path-based entrypoint `Serve.runNode(headConfigPath, privateConfigPath, …)` → `NodeConfig.load(...)`. This is intentional: the test's purpose is an end-to-end "can a node boot from on-disk config files" check.
2. `TestPeers` builds each wallet as a **BloxBean ed25519‑BIP32 extended key** (`WalletModule.BloxBean`, `signExtended`). That 128‑byte extended key cannot be expressed as a 32‑byte vanilla Scalus signing key, so serializing the private config substitutes an **all‑zeros placeholder** (`PeerWallet.dummyPeerWalletEncoder`). `GenerateSampleConfig.writeAll` documents this: *"signing keys are written as dummy all-zero bytes … Replace them before running a node."*
3. The test does **not** replace the placeholder. On load, the wallet is a vanilla `WalletModule.Scalus` wallet with `(real vk, zero sk)`.
4. On startup, `StackComposer.bootstrapInitialStack` composes stack 0, signs its own hard‑acks (`EffectSigner.mkInitialRound1Signatures` → `wallet.mkTxSignature`) over the fallback tx, and `SlowConsensusActor.parseOwn2PhaseAcks` immediately **self‑verifies** them (`HardAckSignatureVerifier.verifyTx` → `platform.verifyEd25519Signature(vk, tx.id, sig)`). Signing with the zero key, verifying under the real vk → `false` **every run**.
5. `BadSignature` is (correctly) treated as an unrecoverable consensus break → `SlowConsensusActor` raises → actor system terminates → `runNode` returns `ExitCode(0)`. `MainSmokeTest` does `IO.race(startedD.get, fiber.join)`; when termination wins the race, the assertion trips.

Neither the terminate‑on‑`BadSignature` behavior nor the test's `ServerStarted` assertion is wrong — the node genuinely cannot sign. The defect is that the generated config files are not bootable.

## Evidence

Instrumenting the sign side (`StackComposer.EffectSigner`) and verify side (`HardAckSignatureVerifier.verifyTx`):

```
SELFSIG-SIGN   txid="a6322565…" vk=6ea31d27… sig=731667ba…
SELFSIG-VERIFY txid="a6322565…" vk=6ea31d27… sig=731667ba… ok=false reverify=false
```

- `txid`, `vk`, `sig` are byte‑identical on both sides (same in‑memory tx, delivered by reference — no persist/replay round‑trip involved).
- Two back‑to‑back verifications both return `false` (`ok=false reverify=false`) → not a racy crypto primitive.
- The written `private.json` literally contains `"signingKey" : "0000000000000000000000000000000000000000000000000000000000000000"`.

## Scope note (why this is bigger than the test)

- `WalletModule.BloxBean` and all mnemonic/HD derivation are **test-only** (confined to `TestPeers`). Production uses `WalletModule.Scalus` (`PeerWallet.scalusWallet`); real keys are minted vanilla via `Bootstrap.generateKeyPair`, and round-trip through the 32-byte codec fine.
- So the "right" fix is to make the test config carry real, round-trippable (vanilla Scalus) keys, and ideally retire the BloxBean wallet module.

**Wrinkle blocking a naive wholesale swap:** `TestPeers` derives keys **deterministically** from `SeedPhrase.Yaci` via BIP32, and the Yaci devnet integration tests (`integration/.../yaci/*`) appear to depend on those derived addresses matching Yaci's **pre-funded genesis UTxOs** (see the `TestPeerName` list capped at 20 with the comment "Stopping here due to Yaci's limit of 20 genesis utxos", and `SeedPhrase.scala`'s notes). Replacing BloxBean/BIP32 with vanilla Scalus keys changes every peer address, which would break Yaci-backed funding. This needs to be confirmed and handled (e.g. keep deterministic-from-seed derivation but produce vanilla keys; and/or decouple Yaci funding from key derivation) before removing BloxBean.

## Suggested fix options

1. **Runnable keys in the test (contained):** give `TestPeers` an opt-in vanilla-Scalus wallet backend (deterministic per `(seed, ordinal)`, default stays BloxBean); `MainSmokeTest` opts in so its `private.json` carries a real, corresponding sk. Unit/mock-backend tests are unaffected; Yaci tests keep BloxBean.
2. **Retire BloxBean wholesale (larger):** switch `TestPeers` to vanilla-Scalus keys everywhere and delete `WalletModule.BloxBean` — **only after** resolving the Yaci genesis-funding dependency above.

Either way, keep the terminate-on-`BadSignature` behavior and the `ServerStarted` assertion as-is.

## Repro

```
sbt "testOnly hydrozoa.app.MainSmokeTest"   # run several times; ~1 in 3–4 fails on main
# inspect the generated config:
grep signingKey /tmp/hydrozoa-smoke-*/config/peer-0/private.json   # → all zeros
```

dragospe:
## Summary

Retires the test-only `WalletModule.BloxBean` wallet backend in favor of deterministic vanilla Scalus keys, and fixes the long-standing `MainSmokeTest` flakiness that turned PRs red (details in #649).

## Why

`TestPeers` derived peer keys as BloxBean ed25519-BIP32 **extended** keys. Those 128-byte extended keys can't be expressed in the 32-byte vanilla-ed25519 codec, so serializing a peer's private config wrote a **dummy all-zero signing key** — the config was not runnable. `MainSmokeTest` boots a node from such a config, so the node signed its stack-0 hard-ack with the zero key, self-verification failed **every run**, and the actor system terminated — racing the HTTP bind the test asserts. The only random part was who won the race, hence the flakiness. (Full root-cause writeup: #649.)

## Changes

1. **`test: retire WalletModule.BloxBean`** — replace BloxBean/BIP32 derivation with a deterministic vanilla Ed25519 keypair per `(seed, ordinal)` and delete `WalletModule.BloxBean`. Vanilla 32-byte keys round-trip through the private-config JSON codec. Production already uses only `WalletModule.Scalus`, so this removes a test-only outlier.
2. **`test(MainSmokeTest): write peer 0's real signing key`** — the stock config encoders deliberately withhold signing keys (`PeerWallet`'s encoder writes a placeholder; real `private.json` files are produced by `GenerateKeyPair` splicing the key into a template). Mirror that: splice peer 0's real signing key into the written `private.json` via a new `TestPeers.deriveScalusKeypair`, so the on-disk config is runnable.

## Testing

- Full unit suite: 303 passed, 0 failed (10 canceled = Yaci-tagged); the changed key values broke no golden tests.
- `just precommit` (fmt/lint/nixfmt) and `just build-werror` pass.
- `MainSmokeTest` run repeatedly: reaches `ServerStarted` every time, with **no** `Hard-ack signature verification failed` logged (previously it fired on every run).

## Notes

- The Yaci integration tests are unaffected: they fund whatever address a peer yields via the devnet faucet (`DevKit.topup` → `/addresses/topup`), not BIP32-derived genesis addresses.
- The stock encoders' deliberate key-withholding is left unchanged; only the test splices the real key explicitly (as `GenerateKeyPair` does for real configs).

Closes #649.

### DIFF
--- a/src/main/scala/hydrozoa/lib/cardano/wallet/WalletModule.scala
+++ b/src/main/scala/hydrozoa/lib/cardano/wallet/WalletModule.scala
@@ -1,53 +1,19 @@
 package hydrozoa.lib.cardano.wallet
 
-import com.bloxbean.cardano.client.crypto.Blake2bUtil
-import com.bloxbean.cardano.client.crypto.api.SigningProvider
-import com.bloxbean.cardano.client.crypto.bip32.key.{HdPrivateKey, HdPublicKey}
-import com.bloxbean.cardano.client.crypto.config.CryptoConfiguration
-import com.bloxbean.cardano.client.transaction.util.TransactionBytes
-import scala.language.implicitConversions
 import scalus.cardano.ledger.{Transaction, VKeyWitness}
 import scalus.crypto.ed25519.{SigningKey as ScalusSigningKey, VerificationKey as ScalusVerificationKey}
 import scalus.uplc.builtin.ByteString
 import scalus.uplc.builtin.JVMPlatformSpecific.signEd25519
 
-/*
-Cardano adopted BIP-32: Hierarchical Deterministic Wallets in the form of Ed25529-BIP32:
-https://github.com/input-output-hk/adrestia/raw/bdf00e4e7791d610d273d227be877bc6dd0dbcfb/user-guide/static/Ed25519_BIP.pdf
-
-Java implementation (copied to BB):
-https://github.com/semuxproject/semux-core/tree/master/src/main/java/org/semux/crypto/bip32
-
-bouncycastle/scalus: doesn't support Ed25529-BIP32, only vanilla Ed25529.
-
-BIP-32 overview:
-
-Seed (32 bytes)
-    ↓ (SHA-512 hash)
-Extended Key (64 bytes)
-    ↓
-[scalar (32) | prefix (32)] - both parts are used during signing, seed has some bits changed.
-
-
-Regular Verification Key (vk):
-  [public_key (32)] = 32 bytes
-    ↓
-Extended Verification Key (xvk):
-  [public_key (32) | chaincode (32)] = 64 bytes
-    ↓
-Extended Signing Key (xsk):
-  [private_key (64) | public_key (32) | chaincode (32)] = 128 bytes
-
-// For signature verification only first 32 bytes of xvk are used:
-val vkeyForVerification = extendedVkey.take(32)  // First 32 bytes
-
-The public key portion is identical whether it's:
- * Standalone (32 bytes)
- * Inside an extended verification key (first 32 of 64 bytes)
- * Inside an extended signing key (bytes 64-95 of 128 bytes)
-
- */
-
+/** A signing backend: the key types plus the Ed25519 signing primitives a
+  * [[hydrozoa.multisig.consensus.peer.PeerWallet]] builds on.
+  *
+  * Cardano verifies witnesses with vanilla Ed25519 over the tx-body hash (`tx.id`), and the only
+  * backend is [[WalletModule.Scalus]] — 32-byte vanilla Ed25519 keys that round-trip through the
+  * private-config JSON codec. (A former BloxBean / Ed25519-BIP32 backend was removed: its 128-byte
+  * extended signing keys cannot be expressed in the 32-byte codec, so serialized configs carried a
+  * dummy key and were not runnable.)
+  */
 trait WalletModule:
 
     type VerificationKey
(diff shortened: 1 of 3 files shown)


GeorgeFlerovsky (approved):

GeorgeFlerovsky:
> Full unit suite: 303 passed, 0 failed (10 canceled = Yaci-tagged); the changed key values broke no golden tests.

Actually, SBT/ScalaCheck are concealing the full range of results (see #647). Luckily, the real numbers are still good:
```
Passed: Total 492, Failed 0` / `succeeded 351, failed 0
```

PR #658, opened 2026-08-21
refactor: thread actor capabilities via an Env given

dragospe:
## What

Thread each multisig/rule-based actor's **capability handles** through a per-actor `Env` exposed as a Scala 3 `given`, instead of passing `config`/`persistence`/`tracer`/`metrics`/effect-handles explicitly through methods. This is ReaderT-over-IO expressed with context parameters (no monad transformer).

The split:
- **`Env` (a `given`) = capability/effect handles**: `config`, `persistence`, `tracer`, `metrics`, and per-actor effect handles (`cardanoBackend`, `l2Ledger`, `l2Screener`, the CardanoLiaison callbacks, ReplayActor's `targets`/`leadsFastBlock`).
- **Explicit params = data**: the request/message, block/stack numbers, and peer identity/topology (`remoteHead`, `coil`, `own`/`peers`/`hubs`/`coils`, …).

For actors with a start barrier, a second tier **`Env.Connected`** (a plain `final class` with `export env.*` + `val connections`) adds the post-barrier `connections`; reactive methods run `using Env.Connected`. Barrier-less components (`RuleBasedActor`, `ReplayActor`) use a single `Env`.

## Notable decision: connection handling standardized on the barrier/given model

The actors previously split into two undocumented styles for holding resolved connections — a `Ref[Option[Connections]]` + `getConnections` cell (flat actors) vs. resolve-once-and-thread (BlockWeaver). This PR **standardizes on the barrier-bound `Env.Connected`**: the `Ref`/`getConnections` cells are removed; `connections` are bound as a `given` in the `PreStart` handler once `pendingConnections` resolves, then the connected handler is `become`-d. This is behavior-preserving — `PreStart` is still the first mailbox message and still blocks on the barrier before any real message.

## What did NOT change

Public constructors / factory signatures, the wiring layer (`spawnCoreActors`, `MrmTracers`, `Connections`/`Handle`/`Targets` types, `Serve`, the regime managers), and `ReplayActor.replay`'s signature + its two call sites. `Env` is built internally in each actor.

## Scope

13 actors: BlockWeaver, FastConsensusActor, SlowConsensusActor, StackComposer, CoilAckSequencer, CardanoLiaison, RequestSequencer, JointLedger, PeerLiaison{HeadToHead,HubToCoil,CoilToHub}, RuleBasedActor, ReplayActor. (+1491 / −992; churn is mostly the `Ref`→`become` dispatch conversions.)

## Verification

- `scalafmtCheckAll` ✓ · `scalafixAll --check` ✓ · `just build-werror` (main + test + integration, `-Werror`) ✓
- `sbt test` → **510 passed, 0 failed**

## Follow-up considered, not included

A shared `HydrozoaActor` trait was evaluated. The barrier lifecycle can't be DRY'd (the `pendingConnections` union type-test erases when `Connections` is a type parameter), `Env.Connected`'s `export env.*` can't be lifted to a generic base (Scala 3 requires a concrete path), and the recovery/`become` step has three genuinely different shapes. The only extractable win was a `trace` helper — deferred as low-value.

---
Draft for review of the pattern before it's applied further. Comments on the `Env`/`Env.Connected` shape and the connection-handling standardization especially welcome.


euonymos:
Verified independently on `peter/env-as-given-actors`: clean compile with zero compiler warnings, and `core/testOnly *` gives **517 passed, 0 failed** (you cite 510 — the base has moved). The churn really is the `Ref`→`become` conversion; I could not find behaviour hiding in it.

One thing the description undersells: framing this as "standardization" is weaker than what it achieves. The substantive change is that **`getConnections`' `IllegalStateException` becomes unrepresentable** — before, every handler could ask for connections and fail at runtime; now `receiveConnected` cannot be entered without them. That is an impossible state made impossible, and it is the argument that should carry the review.

Three notes, none blocking.

## `CoilRelay` is the 14th actor and is still on the old pattern

`CoilRelay.scala:76-83` has the identical shape — same `pendingConnections` union, same `PreStart` self-send, same `Ref[Option[Connections]]` plus a `getConnections` that raises `IllegalStateException`:

```scala
private val connections = Ref.unsafe[IO, Option[CoilRelay.Connections]](None)

private def getConnections: IO[CoilRelay.Connections] =
    connections.get.flatMap(
      _.fold(IO.raiseError(IllegalStateException("CoilRelay missing its connections.")))(IO.pure)
    )
```

It is not in the scope list and the description does not say why. It carries no `config` / `tracer` / `metrics`, so there is no `Env` to build — but the *connection* half applies exactly, and it is the one place the runtime failure this PR eliminates still exists. Either convert it or state the exclusion; otherwise "the `Ref`/`getConnections` cells are removed" is 13-of-14, and whoever opens `CoilRelay` next cannot tell which pattern is canonical.

## `Env` and `Env.Connected` are not substitutable

`Env extends CardanoNetwork.Section`; `Env.Connected` only does `export env.*`, so it is **not** a `Section`. It works today because `given env: Env` stays in class scope and resolution finds it there — but an `Env.Connected` is not self-sufficient: hand one to a free function and `using CardanoNetwork.Section` will not resolve.

Since you asked about the shape specifically: the two tiers reading as "the same thing plus connections" while differing in what they satisfy is the part I would expect to bite, the first time someone extracts a helper that takes only `using Env.Connected`. Either `Connected extends CardanoNetwork.Section` too, or say in the scaladoc that it is a bundle rather than a subtype.

## The pre-`PreStart` invariant is now load-bearing and undocumented

A message arriving before `PreStart` used to fall through `receiveTotal` — harmless for anything that did not need connections. It now raises `Unexpected message received before PreStart`, and `MultisigRegimeManagerBase` escalates every supervisor directive, so that is a node kill rather than a handled message.

Safe as things stand: every actor is spawned before `pendingConnections` completes, so `PreStart` is always mailbox message #1 and nothing can precede it. That is a fact about the current spawn topology, not about the actor — and the actor is now the thing that depends on it. Worth one scaladoc line on the `receive` barrier saying so, so that an actor spawned mid-life later (handed a ref before its own `preStart` runs) fails at review rather than at runtime.

euonymos (changes requested):
I am not sure about the rest of the Claude's findings - but CoilRelay definitely needs to be migrated.

dragospe:
Ah yeah... this one I was actually thinking of closing because its not really buying us a ton. But I forgot that it also bundled in the `.become` for the connected version so its worth it. I'll update.

euonymos:
Merged `origin/main` into this branch (`aa15d1f5`) and ran the full unit suite: **458 succeeded, 0 failed, 13 canceled, 3 ignored, 2 pending**.

## The collision was version skew, not a design conflict

This branch predates `f499cacc`, which changed `PendingConnections` to carry an `Either`. `resolveConnections` here still treated the deferred value as the `Connections` itself, so it stopped compiling the moment main arrived. The fix is one combinator, not a redesign:

```scala
def resolveConnections[C](pending: PendingConnections | C)(project: Connections => C): IO[C] =
    pending match
        case shared: PendingConnections @unchecked =>
            shared.get.flatMap(IO.fromEither).map(project)
        case direct => IO.pure(direct.asInstanceOf[C])

def resolveConnectionsF[C](pending: PendingConnections | C)(project: Connections => IO[C]): IO[C] =
    pending match
        case shared: PendingConnections @unchecked =>
            shared.get.flatMap(IO.fromEither).flatMap(project)
        case direct => IO.pure(direct.asInstanceOf[C])
```

Everything else in the merge is main's work landing under this branch's `Env` / `Env.Connected` shape.

## Three main-side changes carried into this branch's shape

1. **`BlockWeaver` marker bundle.** Main moved recovery onto the boot markers. Kept the `Env.Connected` given, took main's call shape:
   ```scala
   given Env.Connected <- initializeConnections
   recovered <- State.recover(markers.fastBlockMark, markers.softConfirmed)
   ```
   with `def recover(fastBlockMark: Option[BlockNumber], softConfirmed: Option[BlockNumber])(using env: Env.Connected)`.

2. **`ReplayActor` no longer reads L1.** Main moved the first poll out of the actor — `replay` runs inside an uncancelable handler — so `seedFirstPollResults` and `cardanoBackend` left `Env`, which is now `Env(persistence, targets, leadsFastBlock)`. `recoverOwnAck` is gone; the handoff is `inflight <- reconstructInflightHandoff(own, markers.hardConfirmed, hardAckedStack)` and the poll results go out as `targets.blockWeaver ! firstPollResults`.

3. **`RequestSequencer` auto-merge damage.** Git took main's `next = markers.nextRequestNumber` binding and this branch's removal of the `initializeConnections` generator (it moved into the `PreStart` handler here), leaving a `for` comprehension whose first line was an `=`. Both halves are correct; the composition wasn't. Now:
   ```scala
   private def preStartLocal: IO[Unit] = {
       val next = markers.nextRequestNumber
       for {
           _ <- state.seedNextRequestNum(next)
           ...
       } yield ()
   }
   ```

## One place this branch wins

`StackComposer.panic` — this branch's `IO.raiseError(new RuntimeException(msg))` over main's `throw`. Kept, since a raised error stays inside the effect and a thrown one escapes the actor's error channel.

## Four resolutions I'd like you to confirm

1. **`Env` and `Env.Connected` are inseparable.** I checked whether the barrier-bound actors could collapse to one `Env`: every one of them uses `Env.Connected` more often than plain `Env`. `ReplayActor` is the sole exception (4 plain, 0 connected), and that is exactly because main took the L1 read out of it. So the split stays as you have it.
2. **`pendingConnections.complete(Right(connections))`** kept with this branch's barrier-ordering comment rather than main's phrasing.
3. **`RecoverSeamsTest`'s `L2ParamsMismatch` case** is new on main and applied without conflict, so it kept its pre-merge 6-argument shape and was passing a `Hash32` where `evacuationMapMark: Option[BlockNumber]` now sits. Inserted `None,`.
4. **`Serve.scala`'s `Cf.mkAll`** — resolved to main's block; worth a second pair of eyes since that file took the largest share of the conflict.

The merge is additive (no force-push); rebase on it whenever suits.

https://claude.ai/code/session_01ToXh3J7qij4Hct58tRN92w

euonymos (commented):
I am leaving it up to @GeorgeFlerovsky

PR #665, opened 2026-08-22
feat(consensus): cap the major blocks a stack may cover

euonymos:
First of the slices bounding major-block load on the fast side. This one caps blast radius; it does not on its own stop the spiral (see "What this does not do").

## Config changes

**New head-agreed parameter.** `StackConfig.maxMajorBlocksPerStack`, default **1000**, in a new `stackConfig` section of `HeadParameters` — so it is hashed into the treasury datum and every peer closes and validates stacks under the same ceiling.

**Where it appears:**

| file | change |
|---|---|
| `head-config/head-config.json` | new `headParams.stackConfig.maxMajorBlocksPerStack` |
| `bootstrap/defaults.json` | new `headParams.stackConfig.maxMajorBlocksPerStack`, written by `init-bootstrap-files` / `keygen-fleet` |

```json
"stackConfig" : {
  "maxMajorBlocksPerStack" : 1000
}
```

**Existing configs keep working — no operator action required.** Both decoders inject the default when the field is absent, so a bootstrap directory or head config written before this release still builds and boots:

- `HeadParameters` — for an existing `head-config.json`
- `BootstrapHeadParams` — for an existing `bootstrap/defaults.json`

Both paths are pinned by tests, and verified end to end against a scaffolded head directory: a fresh `keygen-fleet` writes the section, and the same `defaults.json` with the section removed still runs `build-head-config`.

**Why the default is 1000 and not something small.** Hard-confirmation carries a large fixed cost per *stack*, independent of size, so a small cap closes many tiny stacks and *lowers* throughput. Measured on the deposit-bot crash: `confirm ≈ 180s + 0.63s × majors`, giving a sustained drain of 0.41 majors/s at a cap of 100 versus 1.23 at 1000.

## Why

A stack's cost is its **Major** partitions. Each derives a settlement, a fallback, rollouts and refunds, and every peer signs all of them and verifies every remote ack against them. A run of **Minor** blocks collapses into a single partition, so minors are close to free.

`tryCloseAsLeader` drains the *entire* longest contiguous ready prefix, and closure is single-flight. Under a deposit flood every block is Major (`decisions.absorbed.nonEmpty` forces `nextHeaderMajor`), so:

```
many absorbing blocks → every one Major → stack N holds k majors
  → k× sign/verify per peer → hard-confirm slow → more blocks pair into `ready`
  → stack N+1 drains a bigger prefix → k grows → stall
```

Observed on the deposit-bot crash — stacks 38→46 went 45, 157, 322, 1,000, 1,755, 1,567, 1,520, 3,420, then 10,774 and never confirmed. Arrival exceeded drain at every step.

## What it does

**Leader** — `State.longestReadyPrefix(maxMajors)` truncates the prefix before the `cap+1`-th Major. The cut lands *on* a Major, so the last included Major keeps its trailing minors (its partition stays whole) and the remainder opens the next stack with a leading Minor run, which `StackPartition.partition` accepts. `maxMajors` is positive, so the first Major always fits and a truncated prefix is never empty.

**Follower** — a brief whose range covers more majors than the cap is a divergence, alongside the existing structural check: new `MajorCapExceeded` event, warn + panic, handover to the rule-based regime. Head-agreed precisely so a follower can enforce it.

## What this does not do

Slice 1 alone does not stop the spiral — it relocates the divergence from the stack into the `ready` queue. Better (the head keeps hard-confirming) but still unbounded. Bounding production is the next slice.

## Tests

`StackMajorCapTest`, six properties: cap respected; result is a genuine prefix; truncation always cuts on a Major; a within-cap run is taken whole; minors never trigger truncation; a positive cap always admits the first Major. Plus the two config-compatibility tests above.

`sbt "testOnly *"` — 518 passed, 0 failed, 72 suites. `just build-werror` and `just lint` green.

## Note for reviewers

`stackConfig` is a plain value parameter in `HeadParametersGen`, not a `Gen`. Drawing it shifts the seeded random stream that fixed-seed fixtures depend on (`pureApply(Gen.Parameters.default, Seed(0L))`), which silently broke five unrelated suites — `RocksDbL2StoreTest`, `RecoverSeamsTest`, `EutxoL2LedgerRecoveryTest`, `TransientTokensEvacuationPurityTest`, `L2QueryEndpointsTest`. Same trap as `247eef4e` during the request cap. A test wanting a specific cap passes one explicitly.

### DIFF
--- a/src/main/scala/hydrozoa/bootstrap/Bootstrap.scala
+++ b/src/main/scala/hydrozoa/bootstrap/Bootstrap.scala
@@ -13,6 +13,7 @@ import hydrozoa.config.head.multisig.block.BlockConfig
 import hydrozoa.config.head.multisig.fallback.FallbackContingency
 import hydrozoa.config.head.multisig.fallback.FallbackContingency.mkFallbackContingencyWithDefaults
 import hydrozoa.config.head.multisig.settlement.SettlementConfig
+import hydrozoa.config.head.multisig.stack.StackConfig
 import hydrozoa.config.head.multisig.timing.TxTiming
 import hydrozoa.config.head.multisig.timing.TxTiming.BlockTimes.{BlockCreationEndTime, BlockCreationStartTime}
 import hydrozoa.config.head.network.{CardanoNetwork, StandardCardanoNetwork}
@@ -159,6 +160,7 @@ object Bootstrap:
         disputeResolutionConfig: DisputeResolutionConfig,
         settlementConfig: SettlementConfig,
         blockConfig: BlockConfig,
+        stackConfig: StackConfig,
         coilQuorum: Int
     )
 
@@ -337,6 +339,7 @@ object Bootstrap:
           disputeResolutionConfig = bhp.disputeResolutionConfig,
           settlementConfig = bhp.settlementConfig,
           blockConfig = bhp.blockConfig,
+          stackConfig = bhp.stackConfig,
           coilQuorum = bhp.coilQuorum,
           // Placeholder: the L2 params hash is not consumed yet. Hash32 requires 32 bytes, so use
           // a zero hash rather than empty bytes (which fail the length check).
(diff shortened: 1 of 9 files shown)


GeorgeFlerovsky:
Unclear whether this particularly helps. For instance, in Ethan's massed-deposit run that slowed hard-consensus to a crawl (with thousands of deposits), the number of major blocks was ~250.

I want to see how well our other hard-confirm optimizations (#677, #678) perform in production, before considering this as a lever.

euonymos:
Superseded by #704.

PR #668, opened 2026-08-23
config: let a node bind somewhere other than it advertises

Quantumplation:
## Why

A peer's `webSocketAddress` in the shared head config does double duty today: it
is the address the rest of the head dials, **and** the address the node binds
(`Serve.scala`). That works on a flat network and breaks as soon as anything
terminates connections in front of the node.

Concretely: we're putting external coil peers behind an ALB that demands a
client certificate. Coils dial `wss://head.sugar.rush.sundae.fi:4001`, the ALB
verifies the certificate and forwards to the head. The head must therefore
*advertise* the public name and *listen* on an interface the ALB can reach —
and `head.sugar.rush.sundae.fi` is not an address any host can bind.

## What

`peerBindHost` / `peerBindPort` on `NodePrivateConfig`. Both optional and
independent; absent, a node binds whatever it advertises, so **every existing
config behaves exactly as before**.

Private config rather than shared is deliberate: where a node listens is that
machine's business, and putting it in the shared config would force every peer
to agree on a detail none of them can observe or verify.

Resolution is extracted to `Serve.peerBindAddress` — a pure function, so it can
be tested. The previous inline version was reachable only by starting a node.

## Testing

`PeerBindAddressTest` covers: default (bind == advertised), full override,
host-only and port-only override, malformed overrides failing loudly rather
than silently binding elsewhere, and a portless advertised address still erroring
when nothing overrides it.

Full suite on a clean build: 56 suites, 0 aborted, 391 passed. The 7 failures
are `CardanoBackendBlockfrostTest`, which needs a real Blockfrost project token
and fails identically on an unmodified `main`.

### DIFF
--- a/src/main/scala/hydrozoa/app/Serve.scala
+++ b/src/main/scala/hydrozoa/app/Serve.scala
@@ -226,6 +226,53 @@ object Serve {
       * `None` and mounts no L2-query endpoints. The EUTXO ledger owns its own RocksDB persistence,
       * separate from the consensus store.
       */
+    /** Where this node's peer websocket server listens.
+      *
+      * Normally that is the address the shared head config advertises for this peer, so there is
+      * one source of truth and no way for the two to drift. The override exists for the case where
+      * they genuinely differ: with a TLS-terminating proxy in front, peers dial a public name that
+      * resolves to the proxy, and no host can bind that name. The node then advertises the public
+      * address to the head and listens wherever the proxy can reach it.
+      *
+      * Each half overrides independently -- moving the local port while keeping the advertised one
+      * is a normal thing to want, and requiring both would make that awkward.
+      */
+    private[hydrozoa] def peerBindAddress(
+        peerBindHost: Option[String],
+        peerBindPort: Option[String],
+        advertised: Uri,
+        ownHeadNum: HeadPeerNumber
+    ): (Host, Port) = {
+        val host = peerBindHost match {
+            case Some(h) =>
+                Host.fromString(h)
+                    .getOrElse(throw IllegalArgumentException(s"peerBindHost is not a host: $h"))
+            case None =>
+                advertised.host
+                    .flatMap(h => Host.fromString(h.value))
+                    .getOrElse(
+                      throw IllegalArgumentException(
+                        s"own head peer $ownHeadNum webSocketAddress has no valid host: $advertised"
+                      )
+                    )
+        }
+        val port = peerBindPort match {
+            case Some(p) =>
+                p.toIntOption
+                    .flatMap(Port.fromInt)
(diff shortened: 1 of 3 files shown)


GeorgeFlerovsky (approved):
LGTM

GeorgeFlerovsky:
This is inbound. It changes where the server listens.

Meanwhile, `remoteHeadUris` still dials every otherpeer at its advertised address directly. If the motivating deployment is an ALB that demands a client certificate, then a node behind it needs to present one when dialing out too — and nothing here does that.

Is inbound-only sufficient because only the external coil peers sit behind the proxy and they dial out cleanly?

GeorgeFlerovsky:
Own advertised address is no longer validated when both overrides are set. Previously the host and port were always parsed, because they were always used. Now a malformed `webSocketAddress` for this peer slips past startup. Other peers parse it when building their `remoteHeadUris`, so it surfaces — just not here, and not as early.

Not urgent, but we should probably fix that to make things more robust.

Quantumplation:
presenting client certificates is done by the underlying HTTP library; it just requires some JAVA_OPTS command line flags when running the coil to make java aware of the certs so it can present them:

```
  JAVA_OPTS="$JAVA_OPTS \
    -Djavax.net.ssl.keyStore=/etc/hydrozoa/coil-0.p12 \
    -Djavax.net.ssl.keyStoreType=PKCS12 \
    -Djavax.net.ssl.keyStorePassword=<keystore password>"
    ```

GeorgeFlerovsky:
> presenting client certificates is done by the underlying HTTP library; it just requires some JAVA_OPTS command line flags when running the coil to make java aware of the certs so it can present them:
> 
> ```
>   JAVA_OPTS="$JAVA_OPTS \
>     -Djavax.net.ssl.keyStore=/etc/hydrozoa/coil-0.p12 \
>     -Djavax.net.ssl.keyStoreType=PKCS12 \
>     -Djavax.net.ssl.keyStorePassword=<keystore password>"
>     ```
> ```

Confirmed. These JAVA_OPTS do make it into the peer-dial mechanism currently used by Hydrozoa. Specifically, Hydrozoa is currently using `JdkWSClient.simple[IO]` (`Serve.scala:161`). That uses `SSLContext.getDefault()`, so it does get your settings from `JAVA_OPTS`. That's why it's working in your test run.

If we change to a different client lib, we'll need to make sure it gets those settings.

GeorgeFlerovsky:
One caveat to keep in mind for the future, though: those properties you set are JVM-global, which means they'll apply to every outbound TLS connection the node makes -- peer dialing, the screener client, the Blockfrost Cardano backend, the Sugar Rush Ledger WS connection.

It might impact performance on those connections. It may also break some of those connections due to weird certificate authority (CA) interactions.

PR #670, opened 2026-08-23
feat(metrics): report head equity on /head/stats

euonymos:
## Why here

srtop reaches hydrozoa over **HTTP only**. It opens two RocksDBs as read-only secondaries — Sugar Rush's event log and the aggregator's views DB — and never hydrozoa's store (the one place hydrozoa's data dir appears in its config is as a directory to size-watch). So equity cannot be read locally by the dashboard; it has to travel on an endpoint.

`/head/stats` is the right one: srtop already polls it, and its contract is "cheap, served from a snapshot, no storage reads", which a pushed gauge satisfies.

## What it reports

`MultisigTreasuryUtxo` already carries `equity` as a field, so nothing needs deriving — this is the same quantity as the anchor `InitializationTx.Parse` checks:

```
treasury.value == evacuation map total + equity + beacon
```

`StackComposer` threads and rotates that treasury, so it reports:

- after `recoverOrBootstrap` — from the recovered treasury, or the initialization treasury `State.initial` starts from, so the gauge shows the real figure **from boot** rather than zero until the first stack closes
- after each `afterClose`, on both the leader and follower paths — the points where the tracked treasury rotates

`StackComposer` gains a `metrics: PeerMetrics` constructor field, mirroring `SlowConsensusActor`; `MultisigRegimeManagerBase` already had `metrics` in scope at the construction site.

## Surfaces

| where | name |
|---|---|
| `GET /head/stats` | `equityLovelace` |
| `GET /head/metrics` | `hydrozoa_equity_lovelace` |

`docs/api/openapi.yaml` regenerated by its golden test. `docs/spec/peer-stats-endpoint.md` gains rows in both the metrics catalog and the instrumentation-points table.

## Semantics worth a reviewer's eye

This is equity **as of the last stack this peer closed**, not the last hard-confirmed one. The composer rotates its treasury at close, one step ahead of hard-confirmation. That is the head's own working figure and what it will build the next stack from, but if you'd rather the gauge tracked the settled value it belongs on `SlowConsensusActor` beside `onStackConfirmed` instead, extracted from the confirmed stack's produced treasury. Happy to move it.

## Tests

`sbt "testOnly *"` — 517 passed, 0 failed, 72 suites. `just build-werror` and `just lint` green.

### DIFF
--- a/docs/api/openapi.yaml
+++ b/docs/api/openapi.yaml
@@ -1296,6 +1296,7 @@ components:
       - mempoolSize
       - leaderMempoolDrain
       - sequencerHeadroom
+      - equityLovelace
       properties:
         uptimeSeconds:
           type: integer
@@ -1319,6 +1320,9 @@ components:
           type: integer
           format: int64
         sequencerHeadroom:
+          type: integer
+          format: int64
+        equityLovelace:
           type: integer
           format: int64
     RateStats:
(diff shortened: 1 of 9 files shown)


GeorgeFlerovsky (approved):
LGTM

PR #671, opened 2026-08-24
DO NOT MERGE (storage change) tx: emit Alonzo-format auxiliary data instead of Shelley-era Metadata

nau:
## Why

Amaru has a bug that makes it reject transactions whose auxiliary data uses the Shelley-era `Metadata` encoding. Emitting the Alonzo-era format unblocks submitting via Amaru.

## What

- `L2Metadata.asAuxData` and L1 `Metadata.asAuxData` now build `AuxiliaryData.AlonzoFormat(Some(...))` instead of `AuxiliaryData.Metadata(...)`. The L1 change covers all tx types (Initialization, Deposit, Fallback, Refund, Rollout, Settlement, Finalization).
- Parsing reads metadata via `AuxiliaryData.getMetadata`, so all three auxiliary-data formats (Shelley, Shelley-MA, Alonzo) are accepted. The `Parser.parse(md: MD)` overload became `parse(auxData: AuxiliaryData)`; the now-impossible `AuxDataIsNotMetadata` error is removed.
- Tests that re-wrapped metadata into `AuxiliaryData.Metadata` to fit the old overload now pass the aux data directly. The malformed-metadata generator in `test/Generators.scala` keeps the Shelley constructor on purpose (negative case).

## Testing

`sbt "testOnly *DepositTxTest* *InitializationTxSeqTest* *L2MetadataTest*"` passes (6 properties, 100 cases each), including the asAuxData/parse round-trips.

## Related issue

https://github.com/pragma-org/amaru/issues/1277

### DIFF
--- a/src/main/scala/hydrozoa/multisig/ledger/eutxol2/tx/L2Metadata.scala
+++ b/src/main/scala/hydrozoa/multisig/ledger/eutxol2/tx/L2Metadata.scala
@@ -4,7 +4,6 @@ import cats.syntax.all.*
 import hydrozoa.config.head.initialization.InitializationParameters.HeadId
 import hydrozoa.config.head.initialization.InitializationParameters.HeadId.toHex
 import scala.util.Try
-import scalus.cardano.ledger.AuxiliaryData.Metadata as MD
 import scalus.cardano.ledger.{AssetName, AuxiliaryData, Metadatum, MultiAsset, Transaction, Word64}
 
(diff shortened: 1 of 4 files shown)


GeorgeFlerovsky:
Thanks @nau. Very useful, but we have to be careful about when we merge it, since it changes the storage format (we store txs), which could complicate things for the mainnet deployed head. We might do it on the weekend.

PR #672, opened 2026-08-24
fix(transport): fail a silent WS link so the dialer redials

euonymos:
## The bug

A dialer re-dials only when `WsDuplex.run` completes. Nothing made it complete on a **half-open socket**:

- `reader` parks on `receiveStream` — no frame, no error
- `writer` keeps *succeeding*: `conn.send` writes into a buffer the peer will never read, and with a proxy in the path those writes are ACKed at TCP level
- so neither side of `IO.race(writer, reader)` ever finishes

Every timer that could rescue it is reset by the coil's own doomed traffic: the `ResendCurrent` tick keeps TCP keepalive (2h default) and any proxy idle-timer alive. And a coil **runs no server**, so unlike head peers it has no Ember `idleTimeout` catching it from the other side.

The result is an absorbing state: the connection looks up, carries nothing, and only a process restart clears it. That matches an incident on the stand — the head was up and healthy, the coil never re-dialed, and work queued behind it until the coil was restarted by hand.

## The fix

A **30s read deadline** on the reader, fed by *any* inbound frame.

The peer already pings every 10s (`NodeWsServer.defaultKeepAlivePing`), so a **quiet** link is not a **silent** one and no application-level heartbeat is needed. But those pings are only visible on the **low-level** `WSConnection`: `connectHighLevel` auto-replies `Pong` and then *recurses past* the `Ping` without surfacing it (the swallow is unconditional — `respondToPings` only gates the reply), and its `send` accepts `WSDataFrame` only, so it can neither observe nor emit control frames.

So `WsDuplex` moves to `client.connect` and takes on what the high level did — `Pong` replies, echoing `Close`, and defragmenting `Text` by the `last` flag, mirroring http4s' own implementation.

**Both dialers move.** `PeerTransport` shares `WsDuplex` and has the identical blind spot; it is merely better covered because head peers also run servers.

## Making it legible

The incident was undiagnosable from logs, for three separate reasons — all fixed here:

1. **A clean close logged nothing.** When a peer closes gracefully the reader stream ends *normally*, so `once` returns `Unit` and `handleErrorWith` never fires. A coil could disconnect and reconnect every second in total silence. New `DialerDisconnected` event.
2. **`getMessage` is null** for most connection exceptions — 8 of 9 warnings in the incident log read `dialer to hub failed: null`. Now logs the exception class too.
3. **Neither transport had a logger in `logback-docker.xml`** (root `warn`), so `DialerConnected` was suppressed in production. Both registered in **all five** logback configs.

Together these made a wedged coil, a flapping coil and an idle coil produce identical output. They no longer do.

## Tests

`WsDuplexTest`, on `TestControl`'s virtual clock:

- a silent connection fails once the deadline elapses (so the dialer can redial)
- **six pings at 10s hold a quiet link open past the 30s deadline, and are Pong'd** — the case a data-only deadline would wrongly kill
- fragmented text is reassembled into one line before dispatch

`sbt "testOnly *"` — 520 passed, 0 failed, 73 suites. `just build-werror` and `just lint` green.

## Note

`defaultReadIdleTimeout` is 30s against a 10s ping interval — the same 3:1 ratio the server uses (`idleTimeout` 20s / ping 10s), so a single missed ping never drops a healthy link. It is a parameter on `run`, so a caller can tighten it.


src/main/scala/hydrozoa/multisig/consensus/transport/CoilPeerWsTransportEvent.scala
```diff
@@ -32,3 +32,9 @@ object CoilPeerWsTransportEvent:
 
     /** A dialer attempt to the hub failed. */
     final case class DialerFailed(cause: Throwable) extends CoilPeerWsTransportEvent
+
+    /** The connection to the hub ended without error — a clean close, or the read deadline with no
+      * inbound frame. Logged because `once` returning normally is otherwise an unlogged state
+      * change: the dialer goes straight back to redialing and the link looks continuously up.
+      */
+    final case class DialerDisconnected(uri: Uri) extends CoilPeerWsTransportEvent
```
GeorgeFlerovsky: > *The connection to the hub ended without error — a clean close, **or the read deadline with no
> inbound frame**.*

The read deadline does **not** land here. `conn.receive.timeout(readIdleTimeout)` raises
`TimeoutException`, so `once` fails and `handleErrorWith` routes it to `DialerFailed`. This PR's own
test asserts that:

```scala
assert(result.isLeft && result.left.exists(_.isInstanceOf[TimeoutException]))
```

This matters more than a normal doc nit, because separating a wedged link from an idle one is the
stated purpose of the logging half of the PR. An operator who reads that sentence will grep for
`disconnected` after a wedge and find `failed`. Same wording in `CoilPeerWsTransportEvent.scala` and
`PeerTransportEvent.scala` — both need it.

Suggest: *"…ended without error — the peer closed cleanly, or the receive side reached end of
stream. A read-deadline expiry is an error and surfaces as `DialerFailed`."*

src/test/scala/hydrozoa/multisig/consensus/transport/WsDuplexTest.scala
```diff
@@ -0,0 +1,104 @@
...
+import scala.concurrent.TimeoutException
+import scala.concurrent.duration.DurationInt
+
+/** [[WsDuplex]]'s liveness contract, on a virtual clock.
+  *
+  * A dialer re-dials only when `run` completes, so `run` completing is the *only* way a wedged link
+  * is ever recovered. These pin the three outcomes: silence fails, keep-alive traffic alone keeps a
+  * quiet link open, and a close ends it cleanly.
+  */
+class WsDuplexTest extends AnyFunSuite {
```
GeorgeFlerovsky, line 18: `WsDuplexTest`'s scaladoc says it pins three outcomes — *"silence fails, keep-alive traffic alone
keeps a quiet link open, and a close ends it cleanly."* The third test is fragmentation, not close;
there is no close test.

It also can't be written against `scheduled` as it stands. Once `frames` is exhausted:

```scala
case None => IO.never
```

so `receive` never yields `None`, `unNoneTerminate` never terminates normally, and `run` can only
ever end by error. **The one path that produces `DialerDisconnected` cannot be exercised.** Given
that a dialer redials *only* when `run` completes, the normal-completion path is worth a test at
least as much as the timeout path.

Small change: let the schedule carry an explicit end-of-stream, e.g. `List[(FiniteDuration,
Option[WSFrame])]`, with `None` meaning "receiving side closed" and `IO.never` reserved for the
half-open case that test 1 already relies on. Then assert `run` completes with `Right(())` after a
`Close`, and that the close was echoed.

src/main/scala/hydrozoa/multisig/consensus/transport/WsDuplex.scala
```diff
@@ -23,11 +47,27 @@ object WsDuplex {
...
+            Stream
+                .repeatEval(conn.receive.timeout(readIdleTimeout))
+                .unNoneTerminate
+                // The accumulator carries pending `Text` fragments; only a `last` frame flushes it.
+                .evalMapAccumulate(Chain.empty[String]) { (partial, frame) =>
+                    frame match {
+                        case WSFrame.Ping(data) => conn.send(WSFrame.Pong(data)).as((partial, ()))
+                        // Echo the peer's close, as the high-level connection did; the next
+                        // `receive` then yields None and the stream ends.
+                        case close: WSFrame.Close => conn.send(close).as((partial, ()))
```
GeorgeFlerovsky:
```scala
case close: WSFrame.Close => conn.send(close).as((partial, ()))
```

If the peer closed the TCP connection rather than just the WS half, this write can fail — and a
clean close is then reported as `DialerFailed`, undoing the distinction the PR is trying to draw.
`conn.send(close).attempt.void` costs nothing and keeps the classification honest.

src/main/scala/hydrozoa/multisig/consensus/transport/WsDuplex.scala
```diff
@@ -23,11 +47,27 @@ object WsDuplex {
...
         val reader: IO[Unit] =
-            conn.receiveStream
-                .evalMap {
-                    case WSFrame.Text(s, _) => onLine(s)
-                    case _                  => IO.unit
+            Stream
+                .repeatEval(conn.receive.timeout(readIdleTimeout))
+                .unNoneTerminate
+                // The accumulator carries pending `Text` fragments; only a `last` frame flushes it.
+                .evalMapAccumulate(Chain.empty[String]) { (partial, frame) =>
```
GeorgeFlerovsky: `Chain[String]` grows until a `last = true` frame arrives; a peer that never sets it accumulates
without bound. `connectHighLevel` had the same exposure so this isn't a regression, but it is our
code now, and the node this protects just died of `OutOfMemoryError`. A cap that fails the
connection past some sane frame count would fit the PR's own "fail so the dialer redials" logic.

src/main/scala/hydrozoa/multisig/consensus/transport/WsDuplex.scala
```diff
@@ -23,11 +47,27 @@ object WsDuplex {
...
+        // `receive` yields None once the receiving side is closed, ending the stream and so the
+        // race. The deadline rides on each read: any frame — including the peer's keep-alive
+        // `Ping` — resets it, so it measures silence rather than idleness.
         val reader: IO[Unit] =
-            conn.receiveStream
-                .evalMap {
-                    case WSFrame.Text(s, _) => onLine(s)
-                    case _                  => IO.unit
+            Stream
+                .repeatEval(conn.receive.timeout(readIdleTimeout))
```
GeorgeFlerovsky: `Stream.repeatEval(conn.receive.timeout(readIdleTimeout))` allocates a timeout per frame rather than
tracking a deadline across the stream. Almost certainly irrelevant next to the payloads we move, and
I have not measured it — noting it only because the head-mesh lane is our busiest path and we are
currently chasing allocation there. `Stream.timeoutOnPull`-style handling would avoid it if it ever
shows up in a profile.

GeorgeFlerovsky (changes requested):
Looks good, at a high level. But Claude gave some feedback to resolve (i.e. all of the comments below).

src/main/scala/hydrozoa/multisig/consensus/transport/WsDuplex.scala
euonymos: Agreed it's unmeasured, and I'd like to leave it as-is for now — but for a semantic reason rather than just "unprofiled".

`conn.receive.timeout(t)` measures **time since the last frame**, which is exactly the property we want: a deadline on *silence*, reset by any inbound frame including the peer's keep-alive `Ping`. A stream-level deadline measures something subtly different, and the distinction is what makes the design work — a live-but-quiet link (no data frames for minutes, pings every 10s) must not trip it.

That property is now load-bearing: on the two-machine stand the coil sat idle for minutes without false-tripping, then failed 27s after an `iptables -j DROP` cut the path. I'd rather not swap the mechanism for an allocation win without a profile showing it matters, since the replacement would need re-validating against both halves of that behaviour.

Noted as a follow-up gated on the head-mesh profile you're working from — if it shows up there, the fix belongs with the measurement that justifies it.

src/main/scala/hydrozoa/multisig/consensus/transport/CoilPeerWsTransportEvent.scala
euonymos: Fixed in `178c4a6f`. You are right, and the stand run confirms it — the blackhole test logged `dialer to hub failed: TimeoutException: 30 seconds`, with no "disconnected" line anywhere. Reworded in both event files to: *"the peer closed cleanly, or the receive side reached end of stream. A read-deadline expiry is an **error** and surfaces as `DialerFailed`, not here."*

src/main/scala/hydrozoa/multisig/consensus/transport/WsDuplex.scala
euonymos: Fixed in `178c4a6f` — `conn.send(close).attempt.as(...)`, with a comment recording why: a failed echo would report a clean close as `DialerFailed` and undo the distinction this logging exists to draw.

src/main/scala/hydrozoa/multisig/consensus/transport/WsDuplex.scala
euonymos: Fixed in `178c4a6f`. Added `WsDuplex.maxTextFragments = 1024`; exceeding it raises, so the dialer redials — the same "fail rather than hang" logic as the read deadline, and a bounded reconnect beats an unbounded heap. Generous on purpose: the liaison protocols send one line per frame, so hitting it means the peer is misbehaving rather than that a message is large. Covered by a new test.

src/test/scala/hydrozoa/multisig/consensus/transport/WsDuplexTest.scala
euonymos: Fixed in `178c4a6f`, and this was the sharpest of the five — the scaladoc described a close test that did not exist, and as you spotted the harness could not have expressed one.

Schedules now carry `List[(FiniteDuration, Option[WSFrame])]`: `None` is end-of-stream, and running off the end still blocks forever so the half-open case test 1 relies on stays distinct. New test asserts `run` completes with `Right(())` after a `Close` and that the close was echoed — the only path that produces `DialerDisconnected`, and previously unreachable. Scaladoc now says what the tests actually pin.

euonymos:
CI caught a real bug in this PR, now fixed in `7c258851`.

**What broke.** `JdkWSClient.receive` is:

```scala
override def receive = closedDef.tryGet.flatMap {
  case None => F.delay(webSocket.request(1)) *> queue.take.rethrow.map(_.some)
  ...
}
```

`request(1)` **then** `queue.take`. My `conn.receive.timeout(readIdleTimeout)` cancelled the `take` after the demand had already been issued, so every expiry leaked one unit of demand — the JDK delivered a frame nobody consumed into an *unbounded* queue. Each firing compounded it:

```
[coil=0] dialer to hub failed: TimeoutException: 30 seconds
EmberServerBuilder… java.util.concurrent.TimeoutException: 20 seconds
Exception in thread "HttpClient-13-SelectorManager" java.lang.OutOfMemoryError: Java heap space
```

The starved reader also stopped answering pings, which is why the server's own idle timeout appears alongside.

**Why nothing else caught it.** The unit tests use a hand-rolled `WSConnection`, so they never exercise the JDK client's demand model. The two-machine stand didn't either — the deadline never fired there. `Stage4Properties` is the only suite that both runs a real WS transport and makes the deadline fire.

**Fix.** Stop cancelling `receive`. The reader stamps each frame's arrival; a watchdog fiber races it and fails when `now - lastFrame >= readIdleTimeout`, polling at a third of the window. Nothing is cancelled mid-read, so no demand leaks. Behaviour is unchanged and all five `WsDuplexTest` cases still pass.

`just integration-fast` locally: **Errors 0** (was 4), 62 passed, no OOM. `just build-werror` and `just lint` green.

Worth noting this is the second unbounded-buffer problem in this file — @GeorgeFlerovsky's fragment-cap comment was pointing at the same class of issue.

GeorgeFlerovsky (approved):
LGTM :)

PR #673, opened 2026-08-25
feat(metrics): report what the StackComposer is waiting for

euonymos:
> **Stacked on #670** — `StackComposer` only gets its `metrics` handle there. Merge that first; this retargets to `main` afterwards.

## Why

Every path through `tryProgress` except deriving returns `IO.unit` **silently**:

```scala
if !s.previousStackHardConfirmed then IO.unit          // single-flight gate
else if canLeadSlow then tryCloseAsLeader              // longestReadyPrefix == Nil -> IO.unit
else tryCloseAsFollower                                // no brief -> IO.unit
                                                       // brief spans unpaired blocks -> IO.unit
```

The follower's not-yet-covered branch even says *"wait silently"* in a comment. So a composer blocked on a peer and one with nothing to do produce identical output — none. Same class of blind spot as the transport work in #672.

## What it reports

| phase | meaning |
|---|---|
| `WaitingForPreviousHardConfirmation` | the single-flight gate — **both roles**, the check precedes the leader/follower branch. Also covers the interval after this peer closed a stack and awaits its confirmation |
| `WaitingForBlockResultsOrSoftConfirmations` | a block is stack-eligible only once **both** halves land: JointLedger's `BlockResult` and fast consensus's `Block.SoftConfirmed` |
| `WaitingForStackBrief` | follower only |
| `Deriving` | building effects |

Plus **`secondsInPhase`**, measured from entry — re-reporting the same phase does *not* restart the clock, so it reads as time **stuck**, not time since the last event. `tryProgress` fires on every inbound message, so that distinction matters.

While `Deriving`: **`partitionsDone` / `partitionsTotal`**. A stack can hold hundreds of partitions.

## Design notes

**One phase for both missing halves.** `pending` is keyed by block number, so block 5 can be missing its `BlockResult` while block 7 is missing its `SoftConfirmed`. A single-valued gauge cannot hold both, hence one phase rather than two.

**Progress is not sampled.** Each partition costs a KZG commitment over a 32768-point SRS plus settlement/rollout tx building; the gauge update is one atomic write. Sampling would save nothing measurable and lose resolution.

**`Deriving` is set before the fold, not during it.** `mkEffectsRegular` runs synchronously inside a single actor message, so nothing else in the composer executes meanwhile — the phase has to be published before entering, and per-partition progress is a plain write from that same thread.

**The builder stays pure.** `mkEffectsRegular` keeps returning `Either` with no `IO`; it takes `onPartitionDerived` / `onDerivationStarted` as `Unit`-returning callbacks defaulting to no-ops, matching how `PeerMetrics` already works. Existing callers and tests are unaffected.

## Surfaces

- `GET /head/stats` → `composer: { phase, secondsInPhase, partitionsDone, partitionsTotal }`
- `GET /head/metrics` → `hydrozoa_composer_seconds_in_phase`, `_partitions_done`, `_partitions_total`, and `hydrozoa_composer_phase{phase="…"}` as a labelled 0/1 set so a dashboard can select on it

`docs/api/openapi.yaml` regenerated by its golden test.

## Tests

`sbt "testOnly *"` — 517 passed, 0 failed, 72 suites. `just build-werror` and `just lint` green.

### DIFF
--- a/src/main/scala/hydrozoa/multisig/consensus/StackEffectsBuilder.scala
+++ b/src/main/scala/hydrozoa/multisig/consensus/StackEffectsBuilder.scala
@@ -69,6 +69,14 @@ object StackEffectsBuilder {
         initialTreasury: MultisigTreasuryUtxo,
         partitions: NonEmptyList[StackPartition],
         initialEvacuationMap: EvacuationMap,
+        // Called with the running count after each partition is derived, so a stack holding
+        // hundreds reports progress rather than going dark until it finishes. A `Unit` callback,
+        // not `IO`: this fold is pure and the only caller writes to an atomic gauge. Defaults to a
+        // no-op so tests and non-instrumented callers are unaffected.
+        onPartitionDerived: Int => Unit = _ => (),
+        // Called once with the partition count before the fold begins, so `partitionsDone` has a
+        // denominator from the start rather than after the first partition lands.
+        onDerivationStarted: Int => Unit = _ => ()
     ): Either[
       Error,
       (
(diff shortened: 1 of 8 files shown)


src/main/scala/hydrozoa/multisig/metrics/PrometheusFormat.scala
```diff
@@ -144,6 +144,28 @@ object PrometheusFormat:
...
+          s.composer.partitionsDone
+        )
+        gauge(
+          "hydrozoa_composer_partitions_total",
+          "Partitions in the stack being built.",
+          s.composer.partitionsTotal
+        )
+        // The phase itself as a labelled 0/1 set, so a dashboard can select on it.
+        StackComposerPhase.values.foreach { p =>
+            b ++= "# TYPE hydrozoa_composer_phase gauge\n"
```
GeorgeFlerovsky: ⛔ Blocking — `hydrozoa_composer_phase` emits four `# TYPE` lines, which fails the whole scrape

```scala
StackComposerPhase.values.foreach { p =>
    b ++= "# TYPE hydrozoa_composer_phase gauge\n"      // inside the loop
    b ++= s"hydrozoa_composer_phase{phase=\"$p\"} ..."
}
```

Rendered (`PrometheusFormat.render` on the test fixture, verbatim):

```
# TYPE hydrozoa_composer_phase gauge
hydrozoa_composer_phase{phase="WaitingForPreviousHardConfirmation"} 1
# TYPE hydrozoa_composer_phase gauge
hydrozoa_composer_phase{phase="WaitingForBlockResultsOrSoftConfirmations"} 0
# TYPE hydrozoa_composer_phase gauge
hydrozoa_composer_phase{phase="WaitingForStackBrief"} 0
# TYPE hydrozoa_composer_phase gauge
hydrozoa_composer_phase{phase="Deriving"} 0
```

The exposition format allows **at most one** `TYPE` line per metric name. Prometheus's text parser
rejects the second one (`second TYPE line for metric name ...`) and **drops the entire scrape** —
so this does not degrade one panel, it takes every hydrozoa metric off the dashboard at once.
There is also no `# HELP` line, which every other family here has.

The file's own established pattern is one `TYPE` before the loop, samples inside it —
`hydrozoa_peer_requests_total` at line 46 and `hydrozoa_blocks_total` at line 52 both do this.
Hoisting the two `b ++=` lines above `foreach` fixes it.

**This is cheap to make un-repeatable.** The suite has no assertion that a metric name is declared
once, which is why four passing tests did not catch it. Adding one finds it immediately:

```scala
test("each metric name declares HELP and TYPE exactly once"):
    val out = PrometheusFormat.render(sample)
    def dupes(keyword: String): Seq[(String, Int)] =
        out.linesIterator.filter(_.startsWith(s"# $keyword "))
            .map(_.split(" ")(2)).toSeq
            .groupBy(identity).view.mapValues(_.size).filter(_._2 > 1).toSeq.sortBy(_._1)
    val _ = assert(dupes("TYPE").isEmpty, s"repeated TYPE lines: ${dupes("TYPE")}")
    assert(dupes("HELP").isEmpty, s"repeated HELP lines: ${dupes("HELP")}")
```

On this branch: `List(("hydrozoa_composer_phase", 4)) was not empty`. It guards every future
family too, not just this one.

src/main/scala/hydrozoa/multisig/metrics/PeerMetrics.scala
```diff
@@ -71,6 +71,12 @@ final class PeerMetrics private (startedAtMillis: Long, remotePeerNums: Vector[I
     private val seqHeadroom = new AtomicLong(0)
     private val equityLovelace = new AtomicLong(0)
 
+    // ---- stack-composer phase (see StackComposerPhase) ----
+    private val composerPhase = new AtomicReference[StackComposerPhase](StackComposerPhase.Deriving)
+    private val composerPhaseSince = new AtomicLong(startedAtMillis)
```
GeorgeFlerovsky: A just-booted head reports `Deriving`, for its whole uptime

```scala
private val composerPhase = new AtomicReference[StackComposerPhase](StackComposerPhase.Deriving)
private val composerPhaseSince = new AtomicLong(startedAtMillis)
```

`reportPhase` is only reached from `tryProgress`, which fires on inbound `BlockResult` /
`SoftConfirmed`, and `bootstrapInitialStack` goes through `mkEffectsInitial`, which is not
instrumented. So a head that has started but not yet processed a block reports
`phase=Deriving, secondsInPhase=<uptime>` — the one reading an operator is meant to escalate on,
produced by a healthy idle node. On a quiet head it never self-corrects.

Two fixes, either fine: an `Idle`/`Starting` case as the initial value, or seed the phase at
`PreStart` the way #670 seeds equity.

src/main/scala/hydrozoa/multisig/consensus/StackEffectsBuilder.scala
```diff
@@ -162,126 +170,134 @@ object StackEffectsBuilder {
...
         // last-of-partition minor (for its SEC). Other minors in a run only get their diffs
         // applied; no KZG paid. The fold short-circuits on the first treasury-build `Left`.
+        onDerivationStarted(partitions.length)
         val seed = Acc(initialTreasury, initialEvacuationMap)
 
         val folded: Either[Error, Acc] = partitions.toList.foldM(seed) { (acc, p) =>
+            // `effectsReversed` gains exactly one entry per partition, so its size after a step is
+            // the number derived so far.
+            def reportProgress(next: Acc): Acc =
+                onPartitionDerived(next.effectsReversed.size)
```
GeorgeFlerovsky: `effectsReversed.size` is O(n) per partition

```scala
def reportProgress(next: Acc): Acc =
    onPartitionDerived(next.effectsReversed.size)
```

`effectsReversed` is a `List`, so this is O(n) per step and O(n²) over the fold. At the ~300
partitions the scaladoc anticipates that is ~45,000 cons traversals, which is nothing against a KZG
commitment — but the fold already knows the index, so it is free to avoid.

PR #675, opened 2026-08-25
Cap each liaison outbox and serve older entries from the journal

GeorgeFlerovsky:
A `LaneOutbound` outbox is pruned only when the remote's pull cursor moves past an entry — not by
confirmation, not by time, not by memory pressure. So a lane whose remote never pulls holds every
item it has ever carried, for the life of the process.

The mainnet head has one. `head-config.json` names coil peers 0 and 1, one `PeerLiaisonHubToCoil` is
created per **configured** coil, `CoilRelay` fans every population artifact out to all of them, and
**coil 1 has never connected in any of the head's six incarnations** (`HardAck:2` holds 0 keys; there
is no `HubHardAck:1` CF at all). `coilQuorum: 1` means coil 0 alone keeps the head healthy, so
nothing surfaces it.

Measured in a controlled A/B on this branch: the head retains **1.11 bytes of heap per byte relayed**
without the cap, linearly, over 20,000 market-maker-shaped requests — while the connected coil sits
on the head's block at every sample. Evidence, arms and caveats:
https://claude.ai/code/artifact/3e7b54d4-61bf-4b2c-a41b-c1dcee1be044

## How to review

Start with `LaneOutbound.scala`; everything else follows from it. Three things, in order of danger:

1. **`reply`'s branch structure** — the `lastAppended = None` short-circuit and the
   `EvictedButUnservable` condition. This is where a mistake wedges a lane.
2. **`Queue` → `TreeMap[N, T]`** in `append` / `reply`. The pruning predicate is unchanged
   (`dropWhile(_ < cursor)` becomes `rangeFrom(cursor)`); the memory-vs-journal test becomes
   `kept.contains(remoteCursor)`, exact where the queue version compared the head.
3. **Two spec edits** in `persistence-and-crash-recovery.md`, which correct a claim the code did not
   honour.

The rest is two named arguments threaded through three liaisons.

## Rationale

The spec already claims this property (§6):

> Reading from the store on demand (rather than eagerly seeding the whole own production) also
> bounds the in-memory outbox to the live tail.

True of what recovery *seeds*, false of what live production *accumulates*. Corrected here.

Eviction is safe for the reason the same section gives: *"Everything that ever reaches a liaison
outbox is persisted before it is appended (CR4) … the outbox is only a cache."* Verified on both feed
paths — `RequestSequencer` writes the journal before `coilRelay ! …`, and a replicated artifact goes
through `persistInbound` before `dispatch`.

So this is not new behaviour. It is what every lane already does immediately after a restart, made
the steady state — and the journal-serve path is already load-bearing in production: when hydrozoa
restarted at 23:21 UTC on 2026-08-24, `seedHighWater` emptied every outbound lane, coil 0 reconnected
pulling from below the floor, and **130 stacks hard-confirmed in the next 65 minutes**.

**Items, not batches.** The cap could have been counted in `maxPerReply` batches. It cannot:
`maxPerReply` is `peerLiaisonMaxRequestsPerBatch` on the request lane and 1 on every other, so any
batch count either starves the single-item lanes or multiplies the one lane whose payloads dominate
memory. Fixing that needs a floor, and the floor then does all the work. One number in items, floored
at the lane's own `maxPerReply`.

## Risk

**The failure mode is a silently wedged lane** — evict an entry the backing cannot produce, and
`reply` returns an empty batch, the remote never advances, and nothing is logged. Two changes make it
loud:

- The lane records the highest number it has evicted; a pull at or below that mark which the journal
  answers with nothing raises `EvictedButUnservable`. Sound rather than heuristic: an evicted entry
  was appended, everything appended is durable, and `JournalScan.loadFrom` limits its *results*, not
  its scan — so silence means the backing is wrong.
- `PeerLiaisonHeadToHead`'s hub-hard-ack lane returned `List.empty` on a non-hub. Now a raise; a
  non-hub never appends there, so it is unreachable, and making it loud keeps it that way.

Worth knowing what "loud" costs today: a raise out of a liaison goes through the cats-actors
supervisor defect, where the original cause is buried and `/ready` keeps reporting `Active`. Still
the right call — the alternative is a lane that stalls with nothing logged — but it is a reason to
ship the `/ready` fix in the same release.

**`reply` no longer reads the store when nothing has been released.** With `lastAppended = None` the
read was issued and its result discarded by the high-water filter, so every non-hub was doing a store
read per pull on its hub-hard-ack lane for the life of the process. That short-circuit is also what
makes the raise above safe; without it, it would fire on every healthy non-hub.

**Config decode.** `peerLiaisonOutboxCap` is optional. Every deployed `private.json` predates it, and
a config that fails to decode is a node that does not start — the same class as the 2026-08-24
crash-loop. The derived decoder is replaced with a hand-written one that defaults the field.

## Testing

The property the change has to have: **a capped lane answers exactly as an uncapped one** over the
same append/pull sequence, driven through a shared journal stub written before each append (CR4).

Its stub counts reads, and the test asserts the capped lane read *more* than the uncapped one. That
control earned its place twice: the first cursor sequence put both lanes on the journal path equally
— it regressed, which defeats an uncapped floor too — so the equivalence held while proving nothing
about eviction.

Also: a cap below `maxPerReply` is floored; eviction plus a journal serve returns the same entry; a
`ceiling` still bounds a post-eviction serve; an unservable evicted entry raises; a lane with nothing
appended never consults the journal (its stub raises if called).

385 unit tests pass; `integration` and `examples` compile. Beyond the unit tests, on this branch:
`CoilLateJoinTest` passes — a coil with no store joining a running, evicting head ends identical to
the head in all six markers, including a deposit absorbed before it existed.

## Scope

A memory bound on one cache. No change to the wire protocol, consensus, or what any peer receives.

It does not address the ~3× per-request amplification (a `lazy val` hex cache in scalus forced by a
`toString` on the ingress path — separate PR), nor the `maxPerReply = 1` throttle on the block and
stack lanes.

`backfill` is renamed to `serveFromJournal` throughout, and the "hot-load" framing goes with it: with
the cap, reading from the journal is the ordinary path for a lagging remote, not an exceptional
catch-up.

### DIFF
--- a/src/main/scala/hydrozoa/config/node/operation/multisig/NodeOperationMultisigConfig.scala
+++ b/src/main/scala/hydrozoa/config/node/operation/multisig/NodeOperationMultisigConfig.scala
@@ -9,6 +9,7 @@ import scala.concurrent.duration.{DurationInt, FiniteDuration}
 final case class NodeOperationMultisigConfig(
     override val cardanoLiaisonPollingPeriod: FiniteDuration,
     override val peerLiaisonMaxRequestsPerBatch: PositiveInt,
+    override val peerLiaisonOutboxCap: PositiveInt,
     override val peerLiaisonResendInterval: FiniteDuration,
     override val rateLimits: RateLimits
 ) extends NodeOperationMultisigConfig.Section {
@@ -24,6 +25,18 @@ object NodeOperationMultisigConfig {
         def peerLiaisonMaxRequestsPerBatch: PositiveInt =
             nodeOperationMultisigConfig.peerLiaisonMaxRequestsPerBatch
 
+        /** How many items each [[hydrozoa.multisig.consensus.liaison.LaneOutbound]] keeps in
+          * memory, floored at that lane's `maxPerReply`. Above the cap the oldest is evicted; a
+          * pull below the remaining floor is served from the journal instead, which is sound
+          * because nothing reaches a lane before it is durable (CR4) and is already how every lane
+          * serves after a restart.
+          *
+          * Node-local on purpose: it changes only how much this peer caches, never what it sends,
+          * so peers may run different values without diverging.
+          */
+        def peerLiaisonOutboxCap: PositiveInt =
+            nodeOperationMultisigConfig.peerLiaisonOutboxCap
+
         /** How often each [[hydrozoa.multisig.consensus.PeerLiaisonHeadToHead]] re-sends its
           * currently outstanding `GetMsgBatch` to the remote peer, to recover from a stalled
           * request-response chain (e.g. caused by a dropped WS frame). The re-send is idempotent on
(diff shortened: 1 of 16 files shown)


src/main/scala/hydrozoa/config/node/operation/multisig/NodeOperationMultisigConfig.scala
```diff
@@ -35,14 +48,39 @@ object NodeOperationMultisigConfig {
...
+            resendInterval <- c.downField("peerLiaisonResendInterval").as[FiniteDuration]
+            limits <- c.downField("rateLimits").as[RateLimits]
+        } yield NodeOperationMultisigConfig(
+          cardanoLiaisonPollingPeriod = pollingPeriod,
+          peerLiaisonMaxRequestsPerBatch = maxRequestsPerBatch,
+          peerLiaisonOutboxCap = outboxCap.getOrElse(defaultPeerLiaisonOutboxCap),
+          peerLiaisonResendInterval = resendInterval,
+          rateLimits = limits
+        )
+    )
```
dragospe, line 85: Is this intended to be removed in a future version? Should it have a deprecation warning?

dragospe (approved):
This broadly makes sense.

PR #676, opened 2026-08-25
Teach every peer to read the protobuf request form, before any peer writes it

GeorgeFlerovsky:
The `Request` lane already has a canonical binary encoding — `proto/request_record.proto`, which the
Sugar Rush ledger reads too. The mesh does not use it: it sends a second, hand-maintained JSON shape
whose payloads are hex, so every request crosses the wire at twice its size.

This is step one of moving the wire onto that encoding. It adds the encoder and makes the decoder
accept **both** forms. It does not change what any peer sends.

## Why it is split in two

A peer running an older build cannot read the new form. So the fleet has to be able to *read* it
before any peer starts *writing* it, or the first upgraded peer takes the mesh down with it. That
ordering is not negotiable and it is not something a config knob should be able to get wrong, so the
flip is a one-line code change in a later release rather than a setting somebody could turn on for
half a fleet.

What this PR is worth on its own: after it ships everywhere, the flip is safe. Before it ships
everywhere, nothing is.

## The two forms cannot be confused

Discrimination is structural, not a tag or a version field:

| form | JSON |
|---|---|
| emitted today | an **object** — `{"kind":…,"requestId":…,"request":{…}}` |
| protobuf | a **string** — base64 of the same bytes the journal stores |

`Decoder` branches on `c.value.isString`. Nothing can be read as both, so there is no precedence
question and no guessing.

## Scope — this is wire size, not the memory fix

Hex in JSON is two characters per byte; base64 is four per three. So a request crosses at about two
thirds of what it costs now, and a transport queue holds correspondingly less.

⚠️ It is **not** a fix for the mainnet retention. The "outbox full of encoded wire lines"
explanation was tested against the OOM heap dump and refuted — the dump held 287 text arrays, not the
~27,000 that would need. Retention lives in the lanes; the cap is what addresses it, and the memory
evidence is at https://claude.ai/code/artifact/3e7b54d4-61bf-4b2c-a41b-c1dcee1be044

It also does not touch cross-repo compatibility. The Rust ledger reads the **store**, and the store
is unchanged — `persistInbound` re-encodes from the typed object, not from the bytes that arrived,
so the wire form never reaches disk.

## Testing

`UserRequestWireCodecTest` covers both forms round-tripping through the decoder the mesh actually
uses, that neither form can be read as the other, that the protobuf form is under two thirds the
size, and that a malformed payload reports *what* was wrong with the record rather than "expected a
string" — the message someone will be reading during a mesh desync.

`RequestRecordStoreFidelityTest` is the wider gate: `encode(decode(stored)) == stored` over every
`Request` record in a synced production store, not the four golden fixtures. Run against the current
mainnet store: **347,319 records, 0 that re-encoded differently, 0 that failed to decode.** That property is
stronger than "decodes without throwing" — a codec that silently dropped a field would still decode,
and would still be wrong for anything that re-encodes what it read. It cancels rather than fails
when no store is present, so a clone without one is not a broken build; point it at one with
`-Dhydrozoa.store.path=…`.

## How to review

`Codecs.scala` is the whole change: two new named codecs, and one `Decoder.instance` that picks
between them. The rest is tests.

The question worth asking of it: is there any JSON value the old encoder could produce that the new
decoder would route to the protobuf branch? The answer is no — the old encoder always produces an
object — but it is the question the split turns on.


src/test/scala/hydrozoa/multisig/persistence/codec/RequestRecordStoreFidelityTest.scala
```diff
@@ -0,0 +1,111 @@
...
+  */
+class RequestRecordStoreFidelityTest extends AnyFunSuite {
+
+    private val storePath: Path =
+        Paths.get(
+          sys.props.getOrElse(
+            "hydrozoa.store.path",
+            s"${sys.props("user.home")}/hz-mainnet/store/rocksdb"
+          )
+        )
```
dragospe, line 31: What your plan for making sure this doesn't bitrot?

dragospe (approved):
Fine in principle, but two questions:

- (As for #675) Are we intended to keep backward compatibility, and if so, why? If not, should we deprecate the old version?
- Is base64-protobuf-over-json the long term plan, or are we planning on going full-protobuf?

PR #678, opened 2026-08-25
Build the evacuation commitment with an NTT and an MSM

GeorgeFlerovsky:
The commitment path did both halves the slow way, and said so itself.

## What was wrong

`KzgCommitment.mkFinalPoly` multiplied the set's binomials out by schoolbook convolution — its own
scaladoc: *"Uses the schoolbook convolution, which is `O(N^2)`"*. `evalFinalPoly` then evaluated the
polynomial one scalar multiplication at a time, under a standing `TODO: use multi-scalar
multiplication, once we have it in the java-blst`.

We have it. Scalus 1.0.0 is already on the classpath and ships both halves — `Poly.product` builds
`∏(x + aᵢ)` with a binary subproduct tree over an NTT, and `bls12_381_G1_multiScalarMul` is a
Pippenger MSM. `BilinearAccumulatorProver.accumulateG1` is exactly those two composed, so the change
is to call it, fed the head's own SRS.

## Same value, faster

Both paths compute `P(τ)·G1` for the same `P`. `KzgCommitmentTest` pins that: it carries the
schoolbook formulation as an independent oracle and asserts the two agree byte for byte — over a
generated set, and at the sizes straddling both of the library's algorithm switches (32/33
iterative-vs-tree, 255/256/257 naive-vs-NTT). Perturbing the oracle makes every non-empty case fail,
so the equality is a real assertion and not a tautology.

It also runs over the real ones. `KzgCommitmentStoreFidelityTest` commits every evacuation map in a
synced production store both ways and compares byte for byte: **1,816 maps, 0 disagreements**. Every
one of them is 30 entries or fewer, so that is correctness on the shapes mainnet actually signs — not
evidence about large N, which is what the sweep below is for.

Both of those are equivalence properties, and equivalence cannot see a commitment that both
implementations compute one way and the deployed binary computed another. `KzgCommitmentOnChainConformanceTest`
closes that from the other side: every standalone evacuation commitment in the store carries the
block it commits to and the value the head signed at the time, so re-deriving from the stored map at
that block compares against recorded output rather than against a second implementation.
**1,785 signed commitments across 1,775 stacks, 0 not reproduced, 0 unpairable.** Pairing each
commitment with the next block's map instead makes the test fail, so it is sensitive to the map it
derives from.

Both store tests skip when no synced store is present, so they are no-ops in CI.

## Measured

Against the demo-stand head's real evacuation maps, both paths fed that head's own SRS:

| map entries | before | after | speedup |
|------------:|-------:|------:|--------:|
| 422 | 42.3 ms | 20.5 ms | 2.1× |
| 3,930 | 1,542 ms | 187.7 ms | **8.2×** |

The old path scales about `N^1.8`, the new one about `N^1.07`. The difference in exponent is the
result; the factor at any one size is just where you sampled it. Full sweep and the Amdahl
accounting: https://claude.ai/code/artifact/ba57c1d6-4960-46bf-8473-ca8e4b4b9ae9

## Cost and safety

The SRS is converted out of blst's `P1` into scalus's `G1Element` once, on first use, and shared
read-only from there. Converting the whole 32,768-point ladder is a fraction of what reading the
trusted setup already costs, and it is immutable, so concurrent commitments share it safely.

The setup-size bound is unchanged: a set larger than the ladder still fails with the same message,
and there is a test for it.

## Scope

`Membership.mkMembershipProofValidated` builds its proof by committing to the complement set, so it
inherits the speedup without changing.

The two JMH benchmarks that measured the old halves are re-pointed at the new ones
(`ProductPolyBenchmark`, `CommitPolyBenchmark`) so the decomposition stays measurable.

A commitment is built per partition, and partitions come from major blocks, so a head doing more
business builds more commitments over a larger map. At the map's design ceiling of 32,767 entries the
two curves extrapolate to roughly **68 s against 1.8 s** per commitment.

373 unit tests pass.

### DIFF
--- a/src/test/scala/hydrozoa/multisig/ledger/commitment/KzgCommitmentOnChainConformanceTest.scala
+++ b/src/test/scala/hydrozoa/multisig/ledger/commitment/KzgCommitmentOnChainConformanceTest.scala
@@ -1,0 +1,134 @@
+package hydrozoa.multisig.ledger.commitment
+
+import hydrozoa.config.head.network.CardanoNetwork
+import hydrozoa.multisig.ledger.block.BlockNumber
+import hydrozoa.multisig.ledger.joint.EvacuationMap as JointEvacuationMap
+import hydrozoa.multisig.ledger.stack.{PartitionEffects, Stack, StackEffects, StackNumber}
+import hydrozoa.multisig.persistence.StoreKey
+import java.nio.ByteBuffer
+import java.nio.file.{Files, Path, Paths}
+import java.util.ArrayList as JArrayList
+import org.rocksdb.{ColumnFamilyDescriptor, ColumnFamilyHandle, DBOptions, Options, RocksDB}
+import org.scalatest.funsuite.AnyFunSuite
+import scala.jdk.CollectionConverters.*
+
+/** Does the current derivation reproduce the commitments a real head already signed?
+  *
+  * The sibling `KzgCommitmentStoreFidelityTest` checks the fast derivation against the schoolbook
+  * one over the same inputs. That is an equivalence property: it cannot detect a commitment that
+  * both implementations compute the same way and the deployed binary computed differently, which is
+  * the failure that would fork a live head or strand funds behind an unspendable treasury.
+  *
+  * This closes that gap from the other side. Every standalone evacuation commitment in the store
+  * carries the block it commits to and the commitment value the head signed at the time — a value
+  * produced by whatever binary was running then, not by this one. Re-deriving from the stored map
+  * at that block and comparing is a conformance check against recorded output rather than against a
+  * second implementation.
+  *
+  * Skipped when no synced store is present, so it is a no-op in CI.
+  */
+class KzgCommitmentOnChainConformanceTest extends AnyFunSuite {
+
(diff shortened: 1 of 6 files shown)


src/test/scala/hydrozoa/multisig/ledger/commitment/KzgCommitmentOnChainConformanceTest.scala
```diff
@@ -0,0 +1,134 @@
...
+  * Skipped when no synced store is present, so it is a no-op in CI.
+  */
+class KzgCommitmentOnChainConformanceTest extends AnyFunSuite {
+
+    private val storePath: Path = Paths.get(
+      sys.props.getOrElse(
+        "hydrozoa.store.path",
+        s"${sys.props("user.home")}/hz-mainnet/store/rocksdb"
+      )
+    )
```
dragospe, line 37: Likewise -- do we want something in, perhaps, git LFS to be able to run this pre-release in CI? Whats the plan to make sure this doesn't bitrot?

dragospe (commented):
Going to let Ilia take a closer look at this one, since I'm not familiar with the KZG internals

PR #679, opened 2026-08-25
Stop rendering requests to log them, retry a config load, and record the failure that stops a node

GeorgeFlerovsky:
Four operational fixes. None changes consensus, the wire, or persistence. Reviewable in this order —
the first is the only one with measurement behind it, the rest are a few lines each.

## 1. The ingress debug event rendered the request it was describing

```scala
tracer.traceWith(RequestDecoded(path, userRequest.toString))
```

The argument is strict, so it runs on every accepted request whatever the log level — and
`RequestDecoded` is a **debug** event, so in production the string is built and thrown away.

The cost is not the string. Rendering a request calls `toString` on its payloads, and a scalus
`ByteString`'s `toString` is its hex, held in an instance `lazy val`:

```scala
override def toString: String = "\"" + toHex + "\""
@threadUnsafe lazy val toHex: String = bytes.toHex
```

Forcing it welds a string of twice the payload's bytes onto that `ByteString` for as long as it
lives — and this one lives a long time: the same instance goes on to the sequencer, the `Request`
journal, and every outbound lane.

That is the multiplier in the mainnet retention arithmetic: the OOM heap held hex and binary arrays
in 1:1 pairs at exactly twice the size, giving **three bytes retained per payload byte** — 73.3 KB
per request against a 24.5 KB mean payload. Counts and the rest of the memory picture:
https://claude.ai/code/artifact/3e7b54d4-61bf-4b2c-a41b-c1dcee1be044

The event now carries the request's kind and payload size, read off fields that already exist. Its
type no longer has a field a rendered request would fit in, which is what stops this coming back.

⚠️ This fixes the multiplier, not the retention. The lane cap is the other half.

## 2. A config load that could not reach Cardano exited the process

Decoding a config resolves the script reference UTxOs against the backend, so loading it is a network
operation — with no retry. One failed resolve raised, the process exited, and the supervisor
restarted it into the same failure. On 2026-08-24 a loss of egress became **61 restarts in a row**.

Backend reads now retry at 2s, 5s, 15s, 30s, 60s — a little under two minutes, long enough to ride
out a DNS blip and short enough that a wrong config still fails while someone is watching the deploy.

Two things it deliberately does not do: it does not retry a malformed config (re-reading the same
bytes gives the same answer, and the distinction is on the error type, not a guess), and it does not
swallow the failure — after the last wait the original error is returned or the original exception
re-raised.

## 3. A supervised actor's failure was never written down

Every directive escalates, and the escalation path raises before the cause is re-reported — so the
log names that secondary failure, not the original, and the node stops with nothing saying why.
`SupervisionDeciderTest` already established the other half: a supervised failure surfaces nothing on
the event stream either way.

The decider is the last place holding the real throwable, so it logs it there — on its own
`Supervision` route at error level in every `logback.xml`, because it is the only record of why a node
stopped. Synchronous, because a decider is a pure `Throwable => Directive` with no effect context.

## 4. A failed UTxO resolution did not say what the backend answered

The status code is what separates "retry, the backend is having a moment" from "this will never
work", and the response body does not carry it — a proxy's 502 body is often empty. It is in the
message now.

## Not fixed here: `/ready`

`/ready` still reports `Active` on a node whose actor system has fail-stopped, which has been
observed: consensus stopped, the dialer failed once a second, readiness stayed green.

The obvious fix does not work. Watching for termination misses the observed case exactly — that
node's system emergency-stopped *without ever completing termination*, which is why it was still
serving HTTP. Detecting it needs a positive liveness signal: a heartbeat routed through the actor
system, a staleness bound on it, and a decision about whether readiness should gate user traffic on
consensus progress. That last part has a real downside — a `/ready` that flips on a slow period takes
a working node out of rotation — so it wants its own change and its own review.

## Testing

`NodeConfigRetryTest` drives the retry loop with the waits injected: a backend that comes back; one
that stays down (asserting the attempt count is bounded and the final error is the backend's own); a
malformed config, not retried at all; a raised failure that clears; one that does not.

`RequestDecodedEventTest` covers the event's contents for both request kinds.

383 unit tests pass; `integration` and `examples` compile.

### DIFF
--- a/src/main/resources/logback-docker.xml
+++ b/src/main/resources/logback-docker.xml
@@ -48,4 +48,6 @@
     <logger name="com.bloxbean" level="warn"/>
     <logger name="scalus" level="warn"/>
     <logger name="CardanoBackend" level="warn"/>
+    <!-- The only record of why a supervised actor failed; never quiet it. -->
+    <logger name="Supervision" level="error"/>
 </configuration>
--- a/src/main/scala/hydrozoa/lib/logging/Logging.scala
+++ b/src/main/scala/hydrozoa/lib/logging/Logging.scala
@@ -18,4 +18,13 @@ object Logging {
     /** log4cats SLF4J adapter used by [[Slf4jTracer.sink]]. */
     def loggerIO(name: String): Logger[IO] =
         Slf4jLogger.getLoggerFromName[IO](name)
+
+    /** A synchronous logger, for the few places that have no effect context to log in.
+      *
+      * A supervision decider is one: it is a pure `Throwable => Directive` the actor library calls
+      * on the failure path, so there is no `IO` to sequence a log into. Everything else should use
+      * [[loggerIO]] or a tracer.
+      */
+    def loggerSync(name: String): org.slf4j.Logger =
+        org.slf4j.LoggerFactory.getLogger(name)
 }
--- a/src/main/scala/hydrozoa/multisig/server/HydrozoaHttpEvent.scala
+++ b/src/main/scala/hydrozoa/multisig/server/HydrozoaHttpEvent.scala
@@ -31,8 +31,16 @@ object HydrozoaHttpEvent:
     /** Decode-failure history (the circe cursor breadcrumb path), separated for diagnostics. */
     final case class JsonDecodeErrorHistory(path: String, history: String) extends HydrozoaHttpEvent
 
-    /** The request body decoded to a domain object (debug). */
-    final case class RequestDecoded(path: String, decoded: String) extends HydrozoaHttpEvent
+    /** The request body decoded to a domain object (debug).
+      *
+      * Carries the request's shape rather than the request. Rendering a decoded request means
+      * `toString` on its payloads, and a `ByteString`'s `toString` is its hex — which scalus caches
+      * on the instance, so a payload that is later retained is retained at three times its size.
+      * The event is built on every request whatever the log level, so that cost is paid whether or
+      * not anyone is reading debug logs.
+      */
+    final case class RequestDecoded(path: String, kind: String, payloadBytes: Int)
+        extends HydrozoaHttpEvent
 
     /** A request was rejected by screening or backpressure — an expected 400, not a fault, so it
       * carries only the client-facing reason (no exception or stack trace).
--- a/integration/src/test/resources/logback-ci.xml
+++ b/integration/src/test/resources/logback-ci.xml
@@ -23,6 +23,8 @@
         </layout>
     </appender>
     <logger name="Mermaid.PeerLiaison" level="info" additivity="false">
+    <!-- The only record of why a supervised actor failed; never quiet it. -->
+    <logger name="Supervision" level="error"/>
         <appender-ref ref="MERMAID"/>
     </logger>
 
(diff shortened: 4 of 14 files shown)


src/main/resources/logback-docker.xml
dragospe, line 1: The comment is duplicated in 4 places, do you think we should just have a `logback-readme.md` or something instead?

src/main/scala/hydrozoa/lib/logging/Logging.scala
dragospe, line 1: This file was removed previously, and _specifically_ the `loggerSync` was eradicated. We don't need it; `SupervisionStrategy` has effectful hooks that accomplish what you're trying to do. See #681 .

src/main/scala/hydrozoa/multisig/server/HydrozoaHttpEvent.scala
```diff
@@ -31,8 +31,16 @@ object HydrozoaHttpEvent:
...
+    /** The request body decoded to a domain object (debug).
+      *
+      * Carries the request's shape rather than the request. Rendering a decoded request means
+      * `toString` on its payloads, and a `ByteString`'s `toString` is its hex — which scalus caches
+      * on the instance, so a payload that is later retained is retained at three times its size.
+      * The event is built on every request whatever the log level, so that cost is paid whether or
+      * not anyone is reading debug logs.
+      */
+    final case class RequestDecoded(path: String, kind: String, payloadBytes: Int)
+        extends HydrozoaHttpEvent
```
dragospe, line 43: The main issue here is what I flagged when I implemented `ContraTracer`. ContraTracer allows you to skip the entire computation associated with a logging call ("rendering" in this case), while slf4j does it regardless of log level. 

IMO, this is PR papering over the issue. We're reducing the multiplier, but still rendering a smaller projection of the user request _which we shouldn't render at all_ unless the log level actually asks for it...

But, due to the way `humanFormat` is written currently, we'll need to rework some things if we want to fix this once and for all. `humanFormat`s' current idiom assigns the log level itself, so the arguments get executed prior to deciding whether to log. If we wanted log-levels natively in a `Contratracer` wrapper, we'd could either to classify events first (sans content) or tag the events up front. 

(Historically, this was left as a TODO because we wanted a thin wrapper over the existing slf4j. The question would be whether to instantiate the tracer with the `logback.xml`-assigned level directly or do something more complicated. Personally, I would probably advise just papering over the slf4j inadequacy via the smarter contratracer squelching, and leave it there until something else bites.)

dragospe: Another issue is that this, and it's corresponding `(...)EventFormat`, are (may be?) misusing the contratracer. The intention is that we carry the actual `UserRequest` in the Event, and only call `_.toString` in the formatter. 

This is what allows us to pattern match over the contratracer with _other_ sinks (for example, in the tests) -- it basically becomes an event bus, and this is prematurely throwing away information. Throwing away the information should be the sink's (formatters) job.

src/main/scala/hydrozoa/lib/logging/Logging.scala
dragospe: But generally,  this _entire file_ was removed, and this is apparently the fifth time some version of it was brought back.

Can we agree to settle on contratracer and convert the re-introduced call sites to it?

dragospe (dismissed):

euonymos (approved):
LGTM

PR #685, opened 2026-08-25
Defer LogEvent rendering behind a per-level gate

dragospe:
Stacked on #684 (bases on `peter/logging-remove-module`). Second of the logging overhaul; the infra half of the level-aware work.

## Why

The sink rendered every event eagerly and let Logback filter by level *afterward* — so a debug event's message (interpolation, domain `toString`, scalus `ByteString` hex) was built even when debug was off. `ContraTracer` (an arrow port of avieth/contra-tracer) can genuinely **skip** all downstream work when squelched, unlike SLF4J which evaluates its argument regardless of level. Nothing wired that to the log level; this does.

## What

- **`LogEventTyped` splits** into eager `level`/`routingKey` (the gate needs them) and a deferred `render: Eval[Rendered]` (msg + ctx + cause).
- **`LogEvent`'s `msg` is by-name**, captured into the `Eval`. So the interpolation — and any `toString` inside it — is deferred at *every* construction site (`From.*` and direct `LogEvent(...)`), with **no call-site changes**. `Slf4jMsg` carries its message as `Eval` too.
- **`levelGated` wrapper**: `sink = rawEmit.levelGated`, where `levelGated = squelchUnlessM(ev => isEnabled(routingKey, level))` and `isEnabled` is a per-trace `org.slf4j` `isEnabled` check (re-queried so runtime Logback level changes are honored; no logback cast). When the level is disabled, the arrow squelches and `render.value` is never forced — no message/`toString` runs. Behaviour at the *output* is unchanged (Logback already filtered by level); the win is that disabled events do no rendering work.
- **Gate wraps only the slf4j leg**, so `Slf4jTracer.sink.contramap(fmt) |+| capture` capture legs always fire (test included).
- The only `org.slf4j` references are the two lines in `Slf4jTracer.scala`, both `// scalafix:ok DisableSyntax` per #684's rule.

## Tests

- Revives `ContraTracerDemo` (deleted in `598dbaff`) and completes its **Demo 5 laziness proof**: an `AtomicInteger` upstream counter advances through a live sink and **does not advance** through `nullTracer` (the squelch drops the whole arrow).
- `Slf4jTracerTest`: a disabled level forces no rendering (counter stays 0), an enabled level forces it exactly once, and a `|+|` capture sink still fires when the slf4j leg is gated off.

## Follow-up (PR-2b)

Per-call `extra` context is still eager (it's varargs). The next PR — the 21 event conversions to carry raw domain objects — also makes `extra` by-name to close the last deferral gap. The expensive vector (the message) is already deferred here.

`just precommit` and `just build-werror` pass.

### DIFF
--- a/src/main/scala/hydrozoa/lib/logging/Slf4jMsg.scala
+++ b/src/main/scala/hydrozoa/lib/logging/Slf4jMsg.scala
@@ -1,6 +1,6 @@
 package hydrozoa.lib.logging
 
-import cats.Monad
+import cats.{Eval, Monad}
 
 /** A small generic message ADT for call sites that don't merit a typed event of their own —
   * typically app entry points (`hydrozoa.app.*`) and test scaffolding. Production actors and shared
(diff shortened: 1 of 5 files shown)


euonymos:
Found this while reviewing: `Slf4jTracerTest` passes alone but fails under `parallelExecution` — `1 did not equal 0` at line 33, i.e. the render *was* forced for the logger configured `off`. CI has not caught it; the only run here was cancelled at 8m21s.

Cause is the SLF4J initialization window: until the backend is installed, `LoggerFactory` hands out substitute loggers, and a substitute reports every level as enabled for every name — so the gate decision opens regardless of `logback-test.xml`. Run alone the suite initializes the backend itself and never sees it.

Fix in #714, against this branch: a bounded wait for the real factory before both assertions. Full suite goes 432+1-failed to 433/0. Two notes for you there — whether the guard belongs in the test or as a `private[logging]` helper on `Slf4jTracer`, and that it needs the `scalafix:ok` escape hatch your own #684 rule now requires.

PR #687, opened 2026-08-25
[spike] scribe migration — PoC done, verdict: GO (viable)

dragospe:
Time-boxed spike: can scribe replace SLF4J + Logback + log4cats behind the `ContraTracer`?

## Verdict: GO (viable)

Two criteria, both now measured in-repo:

- ✅ **Level API without `org.slf4j`** — `ScribeTracer.sink` gates via `scribe.Logger.get(name).includes(level)`; zero `org.slf4j`/log4cats. The decisive win.
- ✅ **Throughput / losslessness** — the sink uses scribe's **synchronous default handle** (`IO { logger.trace(msg) }`): **0 drops, ~397k ev/s** (no-op writer, 320k-event burst / 32 fibers). For the high-volume trace path, a bounded cats-effect `Queue` + consumer fiber is **0 drops, ~384k ev/s** with backpressure — the idiomatic replacement for Logback's `AsyncAppender`.

## Correction to the earlier HOLD

The first pass concluded **HOLD** on a ~99.9%-drop measurement. That was a **benchmarking error**: it configured scribe's opt-in `AsynchronousLogHandle`, whose drain thread writes **one record per `Thread.sleep(1)`** (~1000 rec/s ceiling, verified against scribe 3.19.0 `handler/AsynchronousLogHandle.scala`). The real sink never uses that handle. Benchmarking the actual paths (synchronous + CE-native async) shows lossless, high throughput. The `AsynchronousLogHandle` case is kept as an **ignored anti-pattern demo** in `ScribeLoadTest`.

`design/logging-scribe.md` carries the full source-verified analysis, the corrected benchmark table, and the remaining pre-migration work (SLF4J-facade for third-party floods, config-profile fidelity).

## Layering note (unchanged)

scribe replaces only the **leaf sink** (`Slf4jTracer.sink`); `ContraTracer` stays as the typed event-bus above it — the test capture-bus and any future telemetry/multi-sink fan-out are untouched by the backend swap.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -230,6 +230,9 @@ lazy val core: Project = (project in file("."))
         // Logging
         "ch.qos.logback" % "logback-classic" % "1.5.18",
         "org.typelevel" %% "log4cats-slf4j" % "2.7.1",
+        // scribe: PoC backend behind ContraTracer (spike — see design/logging-scribe.md).
+        "com.outr" %% "scribe" % "3.19.0",
+        "com.outr" %% "scribe-cats" % "3.19.0",
         // Used for input/output
         "org.scala-lang" %% "toolkit" % "0.7.0",
         // cats
--- a/design/logging-scribe.md
+++ b/design/logging-scribe.md
@@ -164,6 +164,25 @@ Remaining, before a full migration:
    `integration-tests.log`) against current Logback output; verify `additivity=false` / sync-vs-async
    choices reproduce.
 
+## Parked follow-up: the `squelchUnlessM` gate duplicates the backend's own gating
+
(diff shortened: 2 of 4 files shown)


dragospe:
Closing: scribe is technically viable but not worth it — scalus/bloxbean keep SLF4J, so a migration wouldn't shrink the stack (it'd swap logback for scribe-slf4j and add scribe-file/scribe-cats), while forcing us to rebuild logback's async appender and absorb scribe's class-vs-logger-name model mismatch. The real logging wins (deferred-render level gate, raw events, ByteString hex hygiene) are already landed on the SLF4J stack and owe nothing to scribe.

PR #688, opened 2026-08-26
perf(logging): drop the redundant level gate, cache the slf4j logger

dragospe:
Three changes to the `Slf4jTracer` sink hot path, none altering behaviour (the deferred-render gate semantics are preserved — an event at a disabled level still never renders, and capture legs still fire).

## What

1. **Drop the redundant `ContraTracer` gate.** The sink forced `ev.render.value` eagerly, then ran `squelchUnlessM(isEnabled)` whose `isEnabled` (`LoggerFactory.getLogger(name).isXEnabled`) duplicated log4cats' own gating — `contextLog` is `F.delay(if (isEnabledUnsafe()) logging())`. Handing the render to log4cats **by-name** lets its single check gate it, so the gate, its per-emit `isEnabled` `IO`, the `Either`/Kleisli routing, and a redundant `getLogger` all go away. (`squelchUnlessM` stays a general `ContraTracer` combinator; only the sink stops using it.)
2. **Cache the log4cats logger per routing key.** `getLoggerFromName` allocates a fresh wrapper on every call; routing keys are a small fixed set, so memoise in a `ConcurrentHashMap`.
3. **`renderMsg` empty-ctx fast path** — return `r.msg` directly instead of `"" + r.msg`.

## Why now

The gate's laziness overlapped log4cats' built-in gating. Relying on log4cats couples the guarantee to the backend — a con only while scribe was a migration candidate; that's off the table (#687 closed), so log4cats is the committed backend and the coupling is fine.

## Measured (JMH `-prof gc`, same-machine before/after; throwaway bench not committed)

| workload | B/op before → after | Δ | ns/op before → after |
|---|---|---|---|
| disabledCheap | 802 → 578 | −28% | 265 → 126 |
| disabledExpensive | 818 → 594 | −27% | 239 → 127 |
| enabledCheap | 1210 → 658 | −46% | 362 → 179 |
| enabledExpensive | 1810 → 1194 | −34% | 439 → 234 |

Every path drops 220–620 B/op and roughly halves per-emit time.

## Tests

`just build-werror` + `just lint-check` clean; the two gate tests (`disabled forces no rendering`; `capture leg still fires`) pass unchanged.

### DIFF
--- a/src/main/scala/hydrozoa/lib/logging/Slf4jMsg.scala
+++ b/src/main/scala/hydrozoa/lib/logging/Slf4jMsg.scala
@@ -1,6 +1,6 @@
 package hydrozoa.lib.logging
 
-import cats.Monad
+import cats.{Eval, Monad}
 
 /** A small generic message ADT for call sites that don't merit a typed event of their own —
   * typically app entry points (`hydrozoa.app.*`) and test scaffolding. Production actors and shared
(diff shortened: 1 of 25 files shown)


dragospe:
Folded into #685 — the sink is now introduced already-optimized (no ContraTracer gate), so the gate never appears in the stack's history. Same change, same measured win (see #685).

PR #690, opened 2026-08-26
DO NOT MERGE UNTIL AFTER COMPETITION — feat(config): move rateLimits into HeadParameters

euonymos:
> [!WARNING]
> **DO NOT MERGE UNTIL AFTER THE COMPETITION.** This changes the config surface —
> `rateLimits` moves out of every peer's private config. Merging it early would
> invalidate the running head's deployment procedure.

First slice of the `headParamsHash` work. Moves `rateLimits` into the head-agreed
parameters and lands the design doc; **no hash is computed or checked yet.**

## Why

Both knobs gate consensus cadence — `softBlockMinPeriod` on the
`FastConsensusActor → BlockWeaver` lane, `hardStackMinPeriod` on
`SlowConsensusActor → StackComposer`. Peers running different values produce
different blocks, so they are not a per-node concern.

They stay their own section rather than joining `txTiming`: `txTiming` holds L1
transaction validity windows, slot-quantized and consumed by transaction builders,
while rate limits are wall-clock gates that reach no transaction.

## Config changes

**What moves.** The whole `rateLimits` section, from `nodeOperationMultisigConfig`
in each peer's private config to `headParams` in the shared head config.

| file | change |
|---|---|
| `peer-private.json` | drop the `rateLimits` object under `nodeOperationMultisigConfig` |
| `head-config.json` | optionally add `rateLimits` under `headParams` |
| `bootstrap/defaults.json` | optionally add `rateLimits` under `headParams` |

**Do existing configs still decode? Yes, all three.** `HeadParameters` and
`Bootstrap.BootstrapHeadParams` both get hand-written decoders that `getOrElse`
the field to `RateLimits.default` — the same treatment `BlockConfig` already gives
`backpressureCoefficient`, and necessary because `deriveDecoder` ignores case-class
defaults. A private config that still carries `rateLimits` decodes too: circe
ignores unknown fields.

**Both decoders verified end to end**, not just by inspection — the config
round-trip suites pass (`Head config.bootstrap config generates`,
`Head config.full config generates`, `Multi-node config.generates`,
`Configuration Codec Properties.round tripping`,
`GenerateSampleConfig.round-trips through disk`).

**Migration for the release notes.** Operators taking the defaults need do
nothing. Operators who tuned `rateLimits` must move the object from each
`peer-private.json` into `headParams` in the shared head config — and every peer
must then carry the same values, which was never enforced before and will be once
`headParamsHash` lands.

## What else is in here

`design/head-params-hash.md` — the digest that will pin the whole head config in
the multisig treasury datum, its byte layout, and the five checks around it
(initialization tx, store identity stamp, evacuation map digest, `l2ParamsHash`,
per-major-block re-verification). Nothing in this PR implements it; the doc is
here so the `rateLimits` move has its rationale in the tree.

`TODO`s at the three stub sites — `HeadParameters.headParamsHash`,
`InitializationParameters.initialEquityContributionsHash`, and the zero-hash
`l2ParamsHash` placeholder in `Bootstrap` — point at the doc. `HeadConfig` had a
scaladoc `@param l2Params` documenting a parameter the class does not have, under
a name no field uses; removed.

## Not in scope, deliberately

Slices 3 and 5 add a field to `MultisigTreasuryUtxo.Datum`, which changes its
`Data` arity and so **cannot be applied to a running head**. They are held until
after the competition regardless of this PR's timing.

## Testing

`just fmt`, `just lint-check`, and `core/testOnly *` (381 tests, 73 suites, 0
failures). `integration/Test/compile` and `examples/Test/compile` are green;
the integration suites themselves were not run.


euonymos:
Closing: `rateLimits` stays **node-local**.

The test is whether a follower's validation depends on the value. It does not — `rateLimits` bounds what a peer does to itself. A leader that shapes aggressively produces fewer, fatter blocks; one that shapes loosely produces more, thinner ones, and followers rebuild from the leader's brief either way. Nothing diverges, so there is nothing for consensus to enforce.

Three costs of moving it that decided this:

1. **Immutable for the head's life.** The digest is pinned in the treasury datum, so a wrong `blockGateSmoothing` cannot be fixed without re-initializing the head. #704 states the backlog gate's dynamics are unexercised in production — five of the seven knobs would be pinned permanently having never been tuned under load.
2. **Uniform sizing across non-uniform hardware**, i.e. sizing for the weakest head peer.
3. It reverses what #704 shipped for: `rateLimits` was kept node-local so the change deploys as a binary swap against a head whose consensus parameters are already pinned.

The bound that *is* worth agreeing head-wide is a follower-enforceable one — `maxMajorBlocksPerStack`, from the closed #665 — tracked as GUM-326.

#691 is being restacked directly onto `main` with the seven `rateLimits` fields dropped from the preimage. `design/head-params-hash.md` from this branch travels with it.

PR #698, opened 2026-08-27
Serve the read-only HTTP surface on coil peers

GeorgeFlerovsky:
A coil runs its own `StackComposer`, and its hard-ack gates the head's next stack — but it served no HTTP at all, so its composer phase, `/head/stats` and `/metrics` were unreachable. During the staging stall on 2026-08-26 that cost real diagnostic time: with no way to query the coil, its idleness had to be inferred from CPU, and an idle composer looks exactly like a wedged one.

## How to review this

One commit. `HydrozoaRoutes.scala` is where the shape change lives; `Serve.scala` is the wiring. The test worth reading is the coil case in `HeadRequestsEndpointsTest`, because it asserts both halves.

## Change

`requestSequencer` becomes an `Option` on `HydrozoaRoutes` and `HydrozoaServer`. It is `None` on a coil — followers accept no user requests — and the submission and admin-finalization routes are simply not mounted when it is absent. That is the idiom `l2QueryReader` already uses, so the surface is read-only by construction rather than by configuration.

The composer's single-flight gate also gains two events. It had none for receiving `Stack.HardConfirmed`, so a composer parked in `WaitingForPreviousHardConfirmation` gave no way to tell a confirmation that never arrived from one that arrived and was then overwritten. `PreviousStackHardConfirmed` and `SingleFlightGateClosed` make both transitions observable.

## Rationale

**The two mutating endpoints stay at the head and tail of `coreEndpoints`.** A head node's endpoint order is therefore unchanged, and so is its golden `openapi.yaml` — the schema diff for this change is empty, which is the check that the head's public surface did not move.

**The integration harness is updated in this commit rather than a follow-up.** Widening the parameter breaks every caller that passed a handle unwrapped, and one of them is the harness, which lives in `integration` — a separate sbt project that `sbt test` does not compile. Splitting it would leave a commit on this branch that builds under `sbt test` and fails under `sbt integration/Test/compile`.

## Risk

The failure mode this introduces is a head node accidentally constructed with `None`, which would silently serve no submission endpoint. The coil test guards the shape from the other direction — it asserts the mutating routes are gone **and** that reads still serve, so a wholly unmounted router fails rather than passes.

`POST /head/requests` is a 405 on a coil, not a 404: `GET /head/requests` is still mounted there, so the path exists and only the method is gone.

## Testing

**547 passed, 0 failed, 0 errors** on this commit, and `integration/Test/compile` green on it too. The suite also reports **10 canceled and 3 ignored** on every commit of this stack: the canceled ones are the Blockfrost-backed tests, which need an API key this environment does not have. They are pre-existing and identical on the base.

Newly covered: a coil's route set, asserted positively and negatively.

Not covered: a coil serving under real load. That is exercised by the staging deployment rather than by a test.

## Scope

An observability change. Nothing about consensus, and nothing a head node serves, is different.

## Base

`base/pr-stack-2026-08-27` — a branch that merges the three open PRs this stack needs and adds nothing of its own:

| merged | branch                                  | why this stack needs it                                  |
| ------ | --------------------------------------- | -------------------------------------------------------- |
| #683   | `peter/bytestring-hex-hygiene`          | carries the whole #675 -> #683 chain                     |
| #673   | `ilia/stack-composer-phase-metrics`     | the composer tracing here builds on it                   |
| #682   | `peter/rulebased-tick-fiber-idempotent` | required by `fix/fiber-lifecycle` further up — see there |

Once those three are on `main`, retarget this PR at `main` and the merge commit disappears from the diff. #670 and #672 are in the staging RC but are **not** dependencies of anything here.

### DIFF
--- a/integration/src/test/scala/hydrozoa/integration/harness/MultiPeerHeadHarness.scala
+++ b/integration/src/test/scala/hydrozoa/integration/harness/MultiPeerHeadHarness.scala
@@ -1690,7 +1690,9 @@ object MultiPeerHeadHarness:
               nodeConfig.headConfig.headPeerNums.toList.map(_.convert).toVector
             )
             HydrozoaRoutes(
-              requestSequencer,
+              // Some: the harness builds a head node's routes, and only a head mounts the
+              // submission and admin-finalize endpoints.
+              Some(requestSequencer),
               conns.blockWeaver,
               // The harness runs no head lifecycle, so readiness is a constant Active.
               IO.pure(NodeStatus.Active),
--- a/src/main/scala/hydrozoa/app/Serve.scala
+++ b/src/main/scala/hydrozoa/app/Serve.scala
@@ -215,7 +215,14 @@ object Serve {
                       httpExtraTracer
                     )
                 case NodeRun.CoilNode(mrm) =>
-                    runCoilNode(system, mrm)
+                    runCoilNode(
+                      nodeConfig,
+                      system,
+                      mrm,
+                      consensusReader,
+                      metrics,
+                      httpExtraTracer
+                    )
             }
         }
     }
@@ -446,8 +453,9 @@ object Serve {
         } yield NodeRun.HeadNode(mrm, l2QueryReader)
     }
 
-    /** Build the coil-node uplink dialer (no inbound server) and allocate the
-      * [[CoilMultisigRegimeManager]]. A coil peer runs no user-facing HTTP server.
+    /** Build the coil-node uplink dialer (no inbound WebSocket server) and allocate the
+      * [[CoilMultisigRegimeManager]]. The coil's HTTP surface is bound separately, in
+      * `runCoilNode`.
       */
     private def buildCoilNode(
         nodeConfig: NodeConfig,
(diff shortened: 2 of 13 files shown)


euonymos:
`runCoilNode` mirrors the head path's `.use(_ => IO.never).start.void`, and that shape has a second problem beyond the orphaned fiber.

`.start.void` drops the fiber handle, so nothing cancels it and Ember never releases — shutdown skips the graceful drain. It also drops the **error channel**: if the port is already bound, the acquisition failure dies inside a fiber nobody holds. The coil logs `Starting HTTP server (coil: read-only surface)...`, then serves nothing, with no error line.

That is the failure mode this PR exists to remove. A coil silently running without its HTTP surface is indistinguishable from one whose surface is up and idle — the same ambiguity that cost diagnostic time on 2026-08-26, reachable from a plain port conflict.

## Suggested change

```scala
            connections <- mrm.connectionsDeferred.get
            serverConfig = httpServerConfig(nodeConfig)
            httpTracer = Slf4jTracer.sink
                .contramap(HydrozoaHttpEventFormat.humanFormat) |+| httpExtraTracer
            _ <- log.info("Starting HTTP server (coil: read-only surface)...")

            // Scoped to the node's lifetime rather than started into a discarded fiber. Two things
            // follow: the server is released on the way out instead of being left to process exit,
            // and a bind failure surfaces here instead of dying unobserved in a fiber nobody holds.
            exit <- HydrozoaServer
                .create(
                  // None on a coil — this is what removes the mutating routes.
                  connections.requestSequencer,
                  connections.blockWeaver,
                  mrm.nodeStatus.get,
                  consensusReader,
                  // A coil always runs a remote ledger, so no L2-query endpoints.
                  None,
                  nodeConfig.headConfig,
                  serverConfig,
                  metrics,
                  httpTracer,
                )
                .use(_ => system.waitForTermination >> abnormalTermination)
        } yield exit
```

Smaller than the original and one nesting level shallower. Compiles on `feat/coil-http-surface`; `core/testOnly hydrozoa.multisig.server.*` passes 41/41.

## The call this asks you to make

With `use`, a bind failure takes the coil down instead of leaving it running blind:

- **today** — port conflict, the coil keeps doing consensus, invisibly, indefinitely;
- **with this** — port conflict, the coil exits, systemd restarts it, it fails again, and you get a visible crash loop.

A coil's hard-ack gates the head's next stack, so this sits on the consensus-participation path and is worth your call rather than a silent refactor. I lean to the second: it is the argument you make for the limiter in #704 — escalation beats a stall with nothing in the log.

If you would rather HTTP failure stay non-fatal on a coil, holding the handle fixes the orphan and the silence without changing survival:

```scala
            serverFiber <- server.use(_ => IO.never)
                .onError(e => log.error(s"Coil HTTP server failed: $e"))
                .start
            _ <- system.waitForTermination.guarantee(serverFiber.cancel)
```

Either way the same shape is on the head path (~40 lines up) and should move with it, or the two roles drift.

euonymos (changes requested):
I think we should not use `start.void` on both `runHeadNode` and `runCoilNode`.

GeorgeFlerovsky:
@euonymos Fixing this in #700 (pushing it now):
https://github.com/cardano-hydrozoa/hydrozoa/pull/698#issuecomment-5445684628

PR #699, opened 2026-08-27
Drop the two unpaged listing endpoints — an unauthenticated caller can OOM a node

GeorgeFlerovsky:
`GET /head/requests` and `GET /head/blocks` each listed an entire RocksDB column family into a heap `List` with no limit and no offset. The `?type=` and `?peer_number=` filters narrowed the *result* but still scanned everything, so they bounded nothing. A single unauthenticated call was therefore an out-of-memory kill, and that is not hypothetical: one operator `curl` to `/head/requests` killed the demo stand on 2026-08-26.

Closes [GUM-298](https://linear.app/gummiworm-labs/issue/GUM-298).

## How to review this

One commit. Read the removal in `HydrozoaRoutes.scala`, then the two new tests, which pin what is left.

## Change

Both endpoints are removed, along with the now-orphaned `listRequests` helper. The generated `openapi.yaml` drops the two operations accordingly.

Everything that remains is keyed by request id or block number, or is a fixed-size summary. `GET /head/requests/{id}` and `GET /head/blocks/{n}` are untouched.

## Rationale

**Removal rather than a `limit` parameter.** A bare limit still needs the full scan to apply it, so it would bound the response and not the work. Re-mounting these wants a cursor plus a bounded page size, which is a different change with a different review.

Note the two different "absent" answers, both correct. On a head node `GET /head/requests` is a **405**, because `POST /head/requests` is still mounted, so the path exists and only the method is gone. On a coil it is a **404**, because nothing is mounted at that path at all.

## Risk

Anything that was calling these breaks. Nothing in this repo does; `srtop` and the sugar-rush services read the store or the keyed endpoints. Operator scripts are the population worth checking before merge.

## Testing

**546 passed, 0 failed, 0 errors** on this commit — one fewer than the commit below it, because removing the endpoints removes their test. The suite also reports **10 canceled and 3 ignored** on every commit of this stack: the canceled ones are the Blockfrost-backed tests, which need an API key this environment does not have. They are pre-existing and identical on the base. Newly covered: that the keyed endpoints still serve, and that the listing paths answer 405/404 as described — the second half matters, because a test that only asserts "not 200" would pass against a broken router.

## Scope

A denial-of-service fix. It removes API surface and adds none.

## Stacking

Sits on top of `feat/coil-http-surface`, which restructured the same routes file. Landing it first on its own needs a manual rebase — the two conflict in `HydrozoaRoutes.scala` and `HeadRequestsEndpointsTest.scala`.

## Base

`feat/coil-http-surface`.

### DIFF
--- a/docs/api/openapi.yaml
+++ b/docs/api/openapi.yaml
@@ -4,40 +4,6 @@ info:
   version: 0.1.0
 paths:
   /head/requests:
-    get:
-      tags:
-      - Requests
-      description: Requests the head has assigned an id — opaque request id, author
-        peer number, and type (transaction or deposit). Filter with ?type=transaction|deposit
-        and ?peer_number=<n>.
-      operationId: getHeadRequests
-      parameters:
-      - name: type
-        in: query
-        required: false
-        schema:
-          type: string
-      - name: peer_number
-        in: query
-        required: false
-        schema:
-          type: integer
-          format: int32
-      responses:
-        '200':
-          description: ''
-          content:
-            application/json:
-              schema:
-                type: array
-                items:
-                  $ref: '#/components/schemas/RequestSummary'
-        default:
-          description: ''
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/ErrorResponse'
     post:
       tags:
       - Requests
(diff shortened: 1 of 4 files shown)


euonymos (approved):
LGTM though the removal strategy is not really coherent - we remove some things but leave some other dead code without any explanations:
1. `ApiDto.RequestSummaryView` and `ApiDto.mkRequestSummaryView`
2. `ConsensusStoreReader.requestsOf` and tests

I might guess that the intent was to start using then again, but it would be nice to know that for sure.

GeorgeFlerovsky:
@euonymos 
For now, I just want to make it impossible for someone to accidentally crash the head with these endpoints. We can think about paging later.
https://github.com/cardano-hydrozoa/hydrozoa/pull/699#pullrequestreview-5046206065

PR #700, opened 2026-08-27
Stop orphaning long-running fibers on replace, and test the weaver's wakeup path

GeorgeFlerovsky:
Four sites stored a long-running fiber's handle without cancelling whatever was already there, or discarded the handle entirely. #682 fixed this shape once in `RuleBasedActor.startTickTimer`; the same shape was still in all three peer liaisons, and `BlockWeaver` had a variant with no handle at all. On the staging soak the weaver's version reached **27,428 live fibers against an idle floor of 174**. After this change the same load class holds 226 → 264.

## How to review this

Start with `BlockWeaver.scala` — it is where the unbounded case was, and where the interesting design question sits. The three liaison diffs are the same two-line change three times. `FiberLifecycleTest` is a source scanner, so read it once and skim it thereafter.

The paragraph under **Risk** is the one to argue with.

## Change

**Liaisons** (`HeadToHead`, `HubToCoil`, `CoilToHub`): `startResendTimer` stored the resend fiber with `resendFiber.set(Some(fib))`, which drops any fiber already held without cancelling it. The orphan is a `foreverM` that never terminates on its own, so it kept delivering `ResendCurrent` for the life of the process. Now `getAndSet` plus cancel.

**BlockWeaver**: `sleepSendWakeup(sleepDuration).start.void` discarded the handle entirely, so the fiber lived until its own sleep elapsed — `min(depositDecisionWakeup, forcedMajorBlockWakeup) - now`, where the forced-major horizon derives from `minSettlementDuration` and is measured in hours. At a high block rate that leaves (entry rate × sleep duration) fibers alive at once, on the leader only, which matches the head/coil asymmetry the soak showed. The handle now lives in `Connections`, already threaded through every state, and is cancelled on replace, where the block completes, and at `stop()`.

## Rationale

**Why the invariant is kept local to the method rather than relying on lifecycle hooks.** A previous review held that the restart-orphan path is unreachable, because cats-actors runs `preRestart` (whose default calls `postStop`) before a restart. Reading cats-actors 2.1.0, that holds for a normal restart — but `faultRecreate` guards it:

```scala
isFailedFatally.ifM(asyncF.unit, actor.aroundPreRestart(cause, msg))
```

so when the actor is `FailedFatally` — set by `Creation.failActor` on an initialization failure — `preRestart` is skipped and `postStop` never runs. Cancelling on replace inside the method means correctness no longer depends on that ordering at all.

**`RemoteL2Ledger` was audited and is correct**: `drop` and the `Resource` finalizer both cancel `receiveFiber`. `Puller.start` returns `IO[Unit]`, not a fiber, so it is not a leak despite the name.

### The alternative that was rejected, and why

Reverting the weaver's cancellation. It was **deliberately removed once before**, in "Fix: prevent race condition when wakeup" (`e6ca939e`, PR #412), because at the time cancellation was what kept a stale wakeup from forcing a block for the wrong block number, and the cancel raced the send. That commit added the block number to `Wakeup` and filtered at the handler instead.

## Risk

Correctness has lived in that handler guard ever since, and that is exactly what makes cancellation safe to reintroduce: a cancel that loses its race now delivers a superseded `Wakeup`, and the guard refuses it. The scaladoc on `scheduleWakeupFiber` says so, so the next reader knows why the cancellation came back rather than reverting it a second time.

**If that guard is ever removed, this cancellation becomes a correctness hazard rather than a resource fix.**

## Testing

**554 passed, 0 failed, 0 errors** on this commit; `integration/Test/compile` green. The suite also reports **10 canceled and 3 ignored** on every commit of this stack: the canceled ones are the Blockfrost-backed tests, which need an API key this environment does not have. They are pre-existing and identical on the base.

Newly covered — **the forced-wakeup path, which had no test at all.** Both header builders in `BlockWeaverTest` hard-coded `mDepositDecisionWakeupTime = None` and no property ever sent a `Wakeup`, so the weaver's deadman — the thing that absorbs a deposit or replaces a fallback tx on a head receiving no requests — was carried entirely by review. `BlockWeaverWakeupTest` pins three halves:

- the wakeups fire (forced-major with no request at all; a deposit target preempting a distant forced-major; a past-due target firing immediately);
- the guard refuses what it should (stale and future block numbers, past and future distinguished);
- cancellation actually collects fibers, rather than letting the guard hide them.

A control property holds both targets far out and asserts nothing completes, so the completion assertions cannot pass vacuously.

**Writing that last test found a real gap, fixed here.** Cancel-on-replace only ran when the *next* wakeup was armed — but leadership rotates, so after completing a block the peer usually becomes a follower and arms nothing for several blocks, leaving the old fiber to sleep out a term that can be hours before delivering a message the guard silently drops. It is bounded to one stale fiber, so it was never the 27k population, but a fiber outliving its block by hours is still wrong. Two smaller gaps closed with it: the non-positive-delay branch fired the wakeup immediately but armed no new fiber, so it never collected the previous one; and `stop()` left whatever was armed still sleeping.

`FiberLifecycleTest` pins the framework behaviour the fix must not depend on, and scans the source for the unguarded shape, so a new occurrence fails a test rather than waiting to be noticed in production.

Not covered: the 27k population itself. It was reproduced on staging and confirmed by ablation there, not in a test.

## Scope

A resource fix. No message, no ordering and no persisted state changes.

## Base

`fix/unpaged-listing-endpoints`, and through it **#682** (`peter/rulebased-tick-fiber-idempotent`) — which is a harder dependency than it looks. `FiberLifecycleTest` scans the whole source tree for the unguarded shape, so it fails on any tree where *any* occurrence remains, and `RuleBasedActor.startTickTimer` is the one #682 fixes. That is the price of a source-scanning test: it makes this branch depend on every other fix of the same shape, rather than only on its own diff. It is worth paying, because the alternative catches nothing when a fifth occurrence appears — but it means #682 has to land first.

### DIFF
--- a/src/main/scala/hydrozoa/app/Serve.scala
+++ b/src/main/scala/hydrozoa/app/Serve.scala
@@ -545,37 +545,36 @@ object Serve {
             _ <- system.actorOf(mrm, "HeadMultisigRegimeManager")
             _ <- log.info("Hydrozoa node started successfully")
 
-            // Start HTTP server once RequestSequencer is available
-            _ <- mrm.connectionsDeferred.get.flatMap { connections =>
-                val serverConfig = httpServerConfig(nodeConfig)
-                val httpTracer = Slf4jTracer.sink
-                    .contramap(HydrozoaHttpEventFormat.humanFormat) |+| httpExtraTracer
-                log.info("Starting HTTP server...") *>
-                    HydrozoaServer
-                        .create(
-                          // Always present on a head; the `Option` exists for the coil.
-                          Some(
-                            connections.requestSequencer.getOrElse(
-                              sys.error("RequestSequencer required on head peers")
-                            )
-                          ),
-                          connections.blockWeaver,
-                          mrm.nodeStatus.get,
-                          consensusReader,
-                          // Some(reader) for a cardano-eutxo node (mounts GET /l2/cardano-eutxo/...); None for
-                          // a remote-ledger node, which serves no L2-query endpoints.
-                          l2QueryReader,
-                          nodeConfig.headConfig,
-                          serverConfig,
-                          metrics,
-                          httpTracer,
-                        )
-                        .use(_ => IO.never)
-                        .start
-                        .void
-            }
+            // The HTTP server needs the RequestSequencer, which only exists once connections
+            // resolve, so wait for them before binding.
+            connections <- mrm.connectionsDeferred.get
+            _ <- log.info("Starting HTTP server...")
 
-            _ <- system.waitForTermination
+            // `surround`, not `start.void`: the server is bound for exactly as long as the node
+            // waits, and its finalizer runs when the actor system terminates. `runCoilNode` has
+            // the same shape and spells out what the old one leaked.
+            _ <- HydrozoaServer
+                .create(
+                  // Always present on a head; the `Option` exists for the coil.
+                  Some(
+                    connections.requestSequencer.getOrElse(
+                      sys.error("RequestSequencer required on head peers")
+                    )
+                  ),
+                  connections.blockWeaver,
+                  mrm.nodeStatus.get,
+                  consensusReader,
+                  // Some(reader) for a cardano-eutxo node (mounts GET /l2/cardano-eutxo/...); None for
+                  // a remote-ledger node, which serves no L2-query endpoints.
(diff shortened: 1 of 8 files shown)


euonymos (approved):
LGTM

PR #702, opened 2026-08-27
Reconcile the stack composer's single-flight gate against persisted confirmation

GeorgeFlerovsky:
The `StackComposer`'s single-flight gate is opened by exactly one `Stack.HardConfirmed` message — no timeout, no retry, no re-derivation — so one lost delivery stops the head permanently.

That happened to the staging head on 2026-08-26. Composition had finished (`partitionsDone == partitionsTotal`), both peers logged the hard confirmation, durable state was fully consistent, and the composer still sat in `WaitingForPreviousHardConfirmation` for 63 minutes until it was restarted.

## How to review this

Smallest change on the branch and self-contained. `StackComposer.scala` is the whole of it; the two design points below are what the review should turn on.

## Change

After two minutes of waiting, the gate is re-derived from `Markers.recoverHardConfirmed` — the same persisted marker `State.recover` already reads. A restart recovers this state precisely because recovery does not wait for a message; this does the same thing without the restart.

Reconciling logs a WARN, because reaching it means a message was lost and the lane that lost it is worth looking at.

## Rationale

**It is driven by `tryProgress`, which runs on every inbound message, so it needs no timer.** That is deliberate rather than incidental: the failure that loses the message can be one that also stops timers firing — the same incident seized the cats-effect worker pool — and a timeout armed on that runtime dies with everything else it was meant to rescue.

**Firing early is harmless, so the threshold needs no tuning.** If the persisted marker has not advanced there is nothing to do, and the check is a `lastKey` point read. The threshold only has to sit above normal hold times. It is injectable so tests need not wait it out.

### Alternatives rejected

**A timeout in the `Limiter`.** It would pace on the delivery path, which is the path that failed. Durable state stayed consistent throughout the incident, so pacing on state is strictly better information — and a `Limiter` timeout dies inside a pool seizure exactly as any other timer does.

**Retrying the delivery.** That fixes one lane. Re-deriving from persistence fixes every way the message can fail to arrive, including ones not yet observed.

## Risk

The reconcile path opens a gate that a message would otherwise have opened, so a spurious open would let a second composition start while the first is live. It cannot: the marker it reads is only written when a stack is genuinely hard-confirmed, so the gate opens on the same fact, just from a different source.

## Testing

**556 passed, 0 failed, 0 errors** on this commit. The suite also reports **10 canceled and 3 ignored** on every commit of this stack: the canceled ones are the Blockfrost-backed tests, which need an API key this environment does not have. They are pre-existing and identical on the base.

`StackComposerReconciliationTest` covers the positive case and its control. **The positive case was verified by ablation** — with the reconcile call removed it fails — and the control ("gate stays shut when persistence has not advanced") still passes, so the suite cannot go green by measuring nothing.

Not covered: the message loss itself. The upstream cause is a cats-effect runtime defect ([typelevel/cats-effect#4674](https://github.com/typelevel/cats-effect/issues/4674)); this change makes the head survive it rather than preventing it.

## Scope

A liveness fix for one gate. It does not address why the message was lost.

## Base

`fix/fiber-lifecycle`.

### DIFF
--- a/src/main/scala/hydrozoa/multisig/consensus/StackComposer.scala
+++ b/src/main/scala/hydrozoa/multisig/consensus/StackComposer.scala
@@ -23,6 +23,7 @@ import hydrozoa.multisig.metrics.{PeerMetrics, StackComposerPhase}
 import hydrozoa.multisig.persistence.recovery.{BlockResultScan, SoftConfirmationScan}
 import hydrozoa.multisig.persistence.{JournalKey, JournalValue, Markers, Persistence, StoreKey, WriteBatch}
 import scala.annotation.tailrec
+import scala.concurrent.duration.{DurationInt, FiniteDuration}
 import scalus.cardano.ledger.{TransactionHash, Value}
 
 /** Stack composer.
@@ -51,7 +52,12 @@ final case class StackComposer(
     pendingConnections: HeadMultisigRegimeManager.PendingConnections | StackComposer.Connections,
     tracer: ContraTracer[IO, StackComposerEvent],
     persistence: Persistence[IO],
-    metrics: PeerMetrics
+    metrics: PeerMetrics,
+    /** How long to wait for a `Stack.HardConfirmed` before re-deriving the single-flight gate from
+      * persistence. Injectable so tests need not wait it out; see
+      * [[StackComposer.HardConfirmReconcileAfter]].
+      */
+    hardConfirmReconcileAfter: FiniteDuration = StackComposer.HardConfirmReconcileAfter
 ) extends Actor[IO, StackComposer.Request] {
     import StackComposer.*
 
@@ -65,6 +71,11 @@ final case class StackComposer(
     private val connections = Ref.unsafe[IO, Option[Connections]](None)
 
     val state: Ref[IO, State] = Ref.unsafe[IO, State](State.initial(config))
+
+    /** Wall-clock millis at which the composer started waiting for a hard confirmation, or `None`
+      * when it is not waiting. Read by [[reconcileHardConfirmedFromPersistence]].
+      */
+    private val hardConfirmWaitSinceMs = Ref.unsafe[IO, Option[Long]](None)
 
     override def preStart: IO[Unit] = for {
         _ <- context.self ! PreStart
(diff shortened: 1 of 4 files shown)


euonymos:
> Firing early is harmless. If the persisted marker has not advanced there is nothing to do.

The marker advances **before** the send, not after it, so there is a window where it has advanced and the message is still legitimately in flight.

`SlowConsensusActor:391-393`:

```scala
_ <- persistHardConfirmation(stackNum, ...)   // writes Cf.HardConfirmation
_ <- conn.cardanoLiaison  ! hardConfirmed
_ <- conn.stackComposer   ! hardConfirmed     // through the Limiter, held hardStackMinPeriod
```

That ordering is correct (CR4, and it is what makes recovery-from-persistence work at all). But it means that for the whole of the limiter's hold, `Markers.recoverHardConfirmed` already reads `stackNum` — exactly the condition `reconcileHardConfirmedFromPersistence` reads as "the message was lost".

## What happens if it fires in that window

1. Reconcile opens the gate, `tryProgress` closes stack `N+1`, and `afterClose` sets `lastClosedStackNum = N+1`, `previousStackHardConfirmed = false` (`StackComposer.scala:948-953`).
2. The held message is released and `handlePreviousStackHardConfirmed` runs. It has no staleness guard (`StackComposer.scala:199-207`) — unconditionally `withPreviousStackHardConfirmed(N)`, which sets `lastClosedStackNum = N` and reopens the gate.
3. `lastClosedStackNum` has regressed, single-flight is open while `N+1`'s hard-ack is outstanding, and `tryProgress` computes `nextStackNum = N+1` again — closing a **second stack under the same number** from the blocks that have arrived since.

## Reachability — the part I cannot settle from here

It needs the hold to exceed `HardConfirmReconcileAfter` (2 min). The repo disagrees with itself:

- `RateLimits.default` and `peer-private.template.json`: `hardStackMinPeriod = 30s` — under, and what the scaladoc cites.
- `integration/.../stage4/Suite.scala:854`, in this stack: narrowed "from its **3-minute** production value" — over.

`rateLimits` is per-node private config and no deployed config is in the repo, so which the live peers run decides whether this is theoretical or active. You will know.

## Two fixes, independent

**Guard the inbound path.** `handlePreviousStackHardConfirmed` should ignore `stackNum < lastClosedStackNum` rather than regress it. Worth having regardless of this PR — it closes the duplicate-stack-number outcome at its source, and today nothing stops a stale or duplicated `Stack.HardConfirmed` doing the same thing.

**Tie the threshold to the hold.** `HardConfirmReconcileAfter` is a bare `2.minutes` whose scaladoc asserts a relationship to `hardStackMinPeriod` that nothing enforces. Deriving it — `max(2.minutes, hardStackMinPeriod * 2)` or similar — makes the invariant structural. That matters more under #704, which makes lane periods dynamic and re-derived.

The rest of the design holds up: driving from `tryProgress` rather than a timer is the right call given the seizure mode, and re-arming before the read keeps it at one point read per interval.

GeorgeFlerovsky:
Unnecessary, buggy over-defensiveness.

PR #717, opened 2026-09-01
design: block hash — commit to request content, not to positions

euonymos:
Design doc only. No code, no behaviour change.

## The gap

Four layers describe a block and none commits to its content:

| layer | commits to |
|---|---|
| `RequestId` = `(HeadPeerNumber, RequestNumber)` | a position — nothing about the payload |
| `BlockBody` = `List[(RequestId, ValidityFlag)]` | the same positions |
| `BlockHeader` | itself; no body field |
| `SoftAck.headerSignature` | `SignedDigest.Onchain(blockNum, startTime, versionMajor, versionMinor)` |

So the multisig on a soft-confirmed block proves agreement on **four scalars**. It covers neither the body nor four of the header's own fields — `endTime`, `fallbackTxStartTime`, `forcedMajorBlockWakeupTime`, `mDepositDecisionWakeupTime`.

`panicOnMismatchWithExpectedBrief` does compare whole briefs structurally, so divergence surfaces. But it is local, it panics rather than refusing, and — the part no relocation fixes — it compares ids, so two peers holding different bytes under the same `RequestId` compare equal.

## What the doc settles

- **`requestHash`** taken at `RequestId` assignment in `RequestSequencer`, returned to the submitter in the same synchronous reply, riding the existing CR1 durable-before-observable barrier. Excludes the `RequestId`, so the same bytes hash the same everywhere and a submitter can compute it before submitting. The hash never travels — every peer recomputes it from the `UserRequestWithId` already fanned out to the mesh.
- **`blockHash`** over the header plus the ordered `(RequestId, requestHash, ValidityFlag)` body, explicit preimage layout, following `EvacuationMap.digest` and `HeadParamsHash`. A `BlockHeader` field, excluded from its own preimage, recomputed and compared wherever a block is rebuilt — on receipt and on replay. Coil peers check it too, which extends the guarantee head↔coil.
- **`SignedDigest(versionMajor, versionMinor, blockHash)`** — `blockNum` and `startTime` collapse into the hash. The versions stay so a ratchet can *order* two signed statements; the doc says plainly that nothing reads them today and that this is option-keeping, not a current requirement.

## Two things worth knowing

**`UserRequestBody.hash` already exists** — blake2b_256, deposits double-hashed, two pinned vectors in `UserRequestTest`, called from nowhere in production. It needs domain separation first: a transaction whose `l2Payload` is exactly the deposit's 64-byte concatenation collides with it.

**Widening the soft-ack preimage costs no Plutus budget.** Despite the name, `SignedDigest.Onchain` never reaches a script — five references repo-wide, `FromData` decoder never called. The on-chain ratchet reads the SEC's own versions and verifies over `voteRedeemer.sec`. Two stale signposts suggest otherwise and are called out: `DisputeResolutionScript`'s comment claims the multisig covers "the blockHeader field of voteRedeemer" when the code signs `sec`, and `VoteTx`/`RatchetVoteTx`/`RuleBasedActor` type SEC signatures as `BlockHeader.Minor.HeaderSignature`.

## Scope

Out: the HTTP surface (separate PR), stack hashes (separate work item, same construction), and the two stale rule-based comments.

Migration: a running head cannot be upgraded across this — signed bytes and the `Block` journal value both move.

One open question left, a profiling one: whether to memoize `blockHash` on the brief.


euonymos:
Pushed `5bc37cbd` — the preimage blocks now match what #734 builds.

Three corrections, all places where the `blockHash` code block disagreed with the prose around it:

1. **`ValidityFlag` is in the per-request encoding.** Scope item 4 and the layout note both put it in; only the code block left it out. Now `u32(peerNum) || u64(requestNum) || raw(requestHash) || u8(validityFlag)`, with `0` valid / `1` invalid pinned.
2. **The forward header times are non-final only.** `BlockHeader.Final` holds none of them — a final block schedules no fallback, no forced major and no deposit decision — so the block shows them under a `non-final block types only` branch.
3. **Every block type writes all three body lists, absent ones as length zero.** The layout note said absent lists stay out of the preimage while the `Initial` paragraph said they encode as empty. The type tag separates the shapes either way, so the lists are written uniformly and the note now says which is which and why.

Two knock-on prose fixes:

- The sentence under the block claimed *every* field is known at the cut. With the flags in, that is wrong — everything except them is, and they are the whole of what defers the announcement. Taking them out of `BlockBody` makes `blockHash` computable at the cut with no other layout change.
- `variable-length fields length-framed` → `lists behind a u32 count`; this preimage has no length-framed field.

And one addition: the `requestHash` construction is written out. The doc called it a public interface and named the two things to fix without ever giving its bytes, so the layout now sits beside `blockHash`'s, and the client-facing page it points at is named (`docs/user-guide/REQUEST-HASH.md`).

The block-type tag values (`0` Initial … `3` Final) are pinned in the code block too.

PR #731, opened 2026-09-09
fix(consensus): derive block zero's soft-confirmation and header timings

euonymos:
Closes GUM-196.

Block zero skips the fast cycle entirely — `Initial` is not a `BlockType.Next` — so no `SoftConfirmation(0)` is ever written, even though block zero is hard-confirmed. Reads landed on `(soft = None, hard = Some)`.

Fixing that exposed a second problem in the same place: block zero's header was stored field by field, when its creation end time already fixes every one of them.

## What changed

**1. Block zero's soft-confirmation is derived on read, in the consensus reader.** `ConsensusStoreReader.softConfirmedAt` returns the moment and resolves block zero to its creation end time; it replaces `softConfirmation` as the only soft-confirmation accessor, so the raw record — which block zero can never have — is not reachable and a reader written tomorrow gets this for free. Nothing is persisted: a `SoftConfirmation(0)` row would store a value the config already fixes, and would cost a `Block.SoftConfirmed.Initial` variant plus a widened `StoreKey.SoftConfirmation.Value`.

That leaves the fast side exposing a moment where the slow side exposes a whole `hardConfirmation` record — block 0 has a real hard-confirmation, because stack 0 spans exactly it. The asymmetry is accidental and is tracked in GUM-335, together with George's fast-consensus-on-block-0 idea that would remove the derived branch entirely.

**2. The header is its end time and nothing else.** `BlockHeader.Initial` is now `(endTime: BlockCreationEndTime)(using TxTiming)`, and the rest are derived members: `startTime` repeats `endTime` (block zero has no creation window), `fallbackTxStartTime` and `forcedMajorBlockWakeupTime` follow through `TxTiming`, and `mDepositDecisionWakeupTime` is absent because block zero absorbs no deposits. The initialization transaction pins `endTime` through its validity end.

**3. So the JSON is `{"endTime": …}` and there is nothing left to check.** `HeadConfig`'s decoder rebuilds the header from the end time and the head params' tx timing — no stored copies, no equality check, no `DecodingFailure` for a header that disagrees with itself. `BlockBrief`'s encoder and decoder are given separately, because only reading block zero back needs the tx timing.

This is the stronger reading of GUM-196's decision 3. "Recompute and check them on parse" polices an affordance; removing the affordance is better. A field nobody can edit is not a field anyone hand-edits by mistake.

**What does not change:** `GET /head/blocks/0` still reports `fallbackTxStartTime` and both wakeups — `ApiDto.BlockHeaderView.InitialView` is built from the header, so the values are still there to read, at the endpoint where an operator should be reading them.

## Config changes

| file | field | change |
|---|---|---|
| `defaults.json` | `blockZeroStartTime` | removed (`blockZeroEndTime` unchanged) |
| `head-config.json` | `blockBrief.header.startTime` | removed |
| `head-config.json` | `blockBrief.header.fallbackTxStartTime` | removed |
| `head-config.json` | `blockBrief.header.forcedMajorBlockWakeupTime` | removed |
| `head-config.json` | `blockBrief.header.depositDecisionWakeupTime` | removed |

**Existing files of both kinds still decode.** The dropped fields are ignored by both decoders — whatever they hold. Verified end to end rather than argued:

- `BlockHeaderInitialTest`: "a config carrying block zero's derived header fields still decodes; they are ignored" — decodes a `head-config.json` whose four dropped fields all carry a value 10 minutes off from the truth, and asserts the resulting header is the one `endTime` determines.
- `BootstrapMembershipTest`: "a defaults.json carrying blockZeroStartTime still decodes; the field is ignored".

The four `head-config.json` fields sit in the block brief, not in the head params, so neither `HeadParameters` nor `BootstrapHeadParams` decodes them; `blockZeroStartTime` is the one that goes through `BootstrapHeadParams`, and its test above covers it. The removals belong in the release notes of the release that ships them.

## headParamsHash

The preimage covers block zero's header as five terms — `u64(startTime) || u64(endTime) || u64(fallbackTxStartTime) || u64(forcedMajorBlockWakeupTime)` plus the optional wakeup — and all five are unchanged bytes: they are derived now instead of parsed, from `endTime` and the tx timing the preimage already covers. Four of the five are therefore redundant, and they stay because dropping a term would mean a new domain tag and an on-chain break for nothing. `docs/spec/head-params-hash.md` and the preimage site say so.

The head-params-hash stack (#730 → #692 → #693 → #694) is merged, and this branch is rebased on it. A head initialized before this PR computes a different digest than one initialized after, because `startTime` was a configured value and is now `endTime`. Nothing is released, so there is nothing to migrate.

## Testing

`CI=1 just precommit` clean, `sbt "testOnly *"` — **474 succeeded, 0 failed**, 13 canceled (store-gated), 3 ignored, 2 pending. `integration`, `petri`, `examples`, and `benchmark` all compile.

https://claude.ai/code/session_016brF9poxv5ogov86qy4ZWk

### DIFF
--- a/docs/spec/head-params-hash.md
+++ b/docs/spec/head-params-hash.md
@@ -204,12 +204,11 @@ chain and moves with hard forks, so it is not something peers agree on.
 reaches the treasury value; the split decides who is paid what at finalization and is otherwise
 unpinned.
 
-The block-zero timing fields matter unevenly and are all included for that reason. `endTime`
-reaches the initialization transaction's validity end; `fallbackTxStartTime`,
-`forcedMajorBlockWakeupTime`, and `mDepositDecisionWakeupTime` reach no transaction at all, so
-the digest is the only thing that pins them. `startTime` repeats `endTime` — block zero has no
-creation window, so it is derived rather than configured — and stays in the preimage because
-removing a term would mean a new domain tag for nothing.
+The block-zero timing fields are five terms over one value. `endTime` reaches the initialization
+transaction's validity end and fixes the whole header: `startTime` repeats it because block zero
+has no creation window, and `fallbackTxStartTime`, `forcedMajorBlockWakeupTime`, and
+`mDepositDecisionWakeupTime` follow from it through the tx timing this preimage already covers.
+All five stay in because dropping a term would mean a new domain tag for nothing.
 
 `hubHeadPeerNumber` decides which head peer relays a coil peer's hard acknowledgement and how
 many `HubHardAck` journals a recovering coil peer must read. It is not in the native script.
--- a/docs/spec/head-params-hash.md
+++ b/docs/spec/head-params-hash.md
@@ -204,10 +204,12 @@ chain and moves with hard forks, so it is not something peers agree on.
 reaches the treasury value; the split decides who is paid what at finalization and is otherwise
 unpinned.
 
-The block-zero timing fields matter unevenly and are all included for that reason.
-`startTime` and `endTime` reach the initialization transaction's validity end, but
-`fallbackTxStartTime`, `forcedMajorBlockWakeupTime`, and `mDepositDecisionWakeupTime` reach no
-transaction at all.
+The block-zero timing fields matter unevenly and are all included for that reason. `endTime`
+reaches the initialization transaction's validity end; `fallbackTxStartTime`,
+`forcedMajorBlockWakeupTime`, and `mDepositDecisionWakeupTime` reach no transaction at all, so
+the digest is the only thing that pins them. `startTime` repeats `endTime` — block zero has no
+creation window, so it is derived rather than configured — and stays in the preimage because
+removing a term would mean a new domain tag for nothing.
 
 `hubHeadPeerNumber` decides which head peer relays a coil peer's hard acknowledgement and how
 many `HubHardAck` journals a recovering coil peer must read. It is not in the native script.
--- a/src/main/scala/hydrozoa/app/Serve.scala
+++ b/src/main/scala/hydrozoa/app/Serve.scala
@@ -312,9 +312,10 @@ object Serve {
                     )
             }
             // Read-only consensus-store view behind the /head/blocks queries.
-            consensusReader = ConsensusStoreReader.fromPersistence(persistence)(using
-              nodeConfig.headConfig
-            )
+            consensusReader = ConsensusStoreReader.fromPersistence(
+              persistence,
+              nodeConfig.headConfig.initialBlock.blockBrief.endTime
+            )(using nodeConfig.headConfig)
         } yield (nodeConfig, system, nodeRun, consensusReader, metrics)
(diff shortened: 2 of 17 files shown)


euonymos:
> why we don't derive `fallbackTxStartTime`, `forcedMajorBlockWakeupTime` and `mDepositDecisionWakeupTime` for the initial block? The config may contain only `endTime`, which is enough to reconstruct.

You're right, and this PR is a halfway house. It derives them and then compares against stored copies, because I implemented GUM-196's decision 3 literally — "recompute and check them on parse instead of trusting the supplied values". The stronger reading of the same intent is: don't store what cannot legitimately differ. `startTime` already got that treatment here; these three deserve it for the same reason.

## It is a contained change

`BlockBrief.Initial`'s codec has exactly one consumer — `HeadConfig`'s decoder at `HeadConfig.scala:146`. Block zero is config, not a spine entry, so this header is never persisted and never broadcast. Dropping the three fields from `blockHeaderInitialEncoder`/`Decoder` touches that one path.

Three things that do **not** break:

- **`headParamsHash` is unaffected.** The preimage covers the header *object*, which still carries all four values — they are derived instead of parsed. Same bytes, same digest, no domain-tag change.
- **The API still shows them.** `ApiDto.BlockHeaderView.InitialView` is a separate DTO built from the header, so `GET /head/blocks/0` keeps reporting `fallbackTxStartTime` and the wakeups. Auditability moves to the endpoint, which is where an operator should be reading them anyway.
- **Existing configs still decode.** circe ignores unknown fields, and this PR already proves the derived values equal what those files carry (`BlockHeaderInitialTest`, "a config written with a block-zero startTime still decodes"). The same test shape extends to the other three.

## The one real trade-off

Carrying-and-checking makes a hand-edit *loud*: change `fallbackTxStartTime` in `head-config.json` and the node refuses to start, naming the field. Derive-only makes it *silent* — the field is ignored and the operator's edit does nothing.

I think silent is still better here, because derive-only removes the affordance rather than policing it. A field nobody can edit is not a field anyone hand-edits by mistake. The loud error only helps someone who already believed the value was configurable, and the fix is to stop presenting it as configuration.

If you want both, the cheap version is to reject the *keys* rather than compare the values: fail decoding if `blockBrief.header` carries any of the four, with a message saying block zero's timings derive from `endTime`. That is one `if` over `c.keys`, keeps the loud failure for hand-edits, and still stores nothing — but it breaks existing configs, which the plain drop does not. I would not do it for that reason.

## Proposal

Drop all three from the header's JSON, leaving `blockBrief.header` as `{ "endTime": ... }`, and delete the comparison in the `HeadConfig` decoder — `BlockHeader.Initial.derive` becomes the only definition, with nothing to disagree with it. That also removes this PR's third config-change row, since the fields simply stop being read.

Say the word and I'll fold it in here rather than leaving it as a follow-up; it is a small delta on top of what is already green.

PR #742, opened 2026-09-14
feat(consensus): drop deposits from an upfront-known final block

euonymos:
Closes part 2 of GUM-190 — "if it's an upfront-known final block, the weaver can ignore register deposit commands".

## What

No later block can absorb a deposit registered in the final block. Registering it anyway costs a `DepositRefundTxSeq.Parse` of the deposit tx and a post-dated refund tx, for a deposit that gets refunded regardless. A leader that knows the block is final *before it opens* drops them instead.

Three sites, all leader-side:

| state | knows because | on a deposit |
|---|---|---|
| `Leader.ProcessingReadyRequests` | `finalizationLocallyTriggered` | prunes pending deposits from the mempool it weaves from |
| `Leader.AwaitingConfirmation` | `finalizationLocallyTriggered` | drops the arriving request |
| `Leader.AwaitingRequest` | that, or `previousBlockConfirmed.finalizationRequested` | drops it **and leaves the block unopened** |

`ProcessingReadyRequests` prunes rather than holds back: held-back overflow would leave the mempool non-empty below `maxRequestsPerBlock`, which the cap check in `AwaitingConfirmation` reads as "nothing held back" and would let a later request from the same peer jump a deposit still sitting in the mempool.

A `finalizationRequested` carried by another peer deliberately does not count in `AwaitingConfirmation` — it only arrives on the confirmation that completes the block, by which point the block's requests are already applied.

## Why it needs no agreement between peers

The decision is the leader's alone. A follower forwards only the requests the brief names, so a deposit the leader dropped never reaches the follower's `JointLedger` either and both still reach the same `blockHash`. The alternative — flagging it `Invalid` in the block — cannot work: `JointLedger` would have to re-derive the verdict on every peer, and a block can flip to final mid-flight (a leader opens block N and forwards requests before N-1 is confirmed), so a follower seeing `BlockBrief.Final` with a registered deposit would diverge and `panicOnMismatchWithExpectedBrief` would take it down.

## What the submitter sees

`UNPROCESSED` on `GET /head/requests/{id}`, permanently. That is not a new outcome: nothing on the submit path knows the head is finalizing (`POST /api/admin/finalize` only signals the `BlockWeaver`; the `RequestSequencer` gates on screening and backpressure alone), and every final-block path already discards the weaver's mempool — `completeBlockFinal`, `Leader.AwaitingRequest.completeBlock`, and both follower `RetiredOnFinalBlock` exits. Anything left in a mempool when the head finalizes already ends up `UNPROCESSED`.

## Config changes

None. No config field, no store format change, no wire change, no block-header or brief change — so no effect on `blockHash`.

## Tests

- `MempoolTest`: `dropDepositRequests` removes exactly the deposits and leaves both sides in arrival order.
- `BlockWeaverTest`: Carol, armed to lead an upfront-known final block, drops a deposit without opening the block or feeding the joint ledger, traces `DepositDroppedFromFinalBlock`, and then still weaves an arriving transaction into `CompleteBlockFinal`.
- `testOnly BlockWeaverTest MempoolTest` — 21 passed, 0 failed.

## Notes

Independent of #741 (part 1 of GUM-190); different code paths, merges in either order.


euonymos:
Closing — not worth carrying.

The saving is a `DepositRefundTxSeq.Parse` and a post-dated refund tx for a deposit that gets refunded regardless. The submitter-visible outcome is unchanged either way: a deposit registered against a final block already ends `UNPROCESSED`, via the mempool discard in `completeBlockFinal`, `Leader.AwaitingRequest.completeBlock`, and both follower `RetiredOnFinalBlock` exits.

Against that, it rewrites three leader states in `BlockWeaver` that #741 also rewrites, so it would cost a conflict resolution on the GUM-190 work that is landing.

Part 2 of GUM-190 is dropped; part 1 lands as #741. Branch `ilia/gum-190-final-block-deposits` is kept if we want to revisit.

PR #749, opened 2026-09-16
feat(transport): prove a liaison peer's identity in the handshake

euonymos:
Closes GUM-322. **Stacked on #744** — merge that first.

The authentication half of GUM-322: steps 1 and 5 of its fix shape, plus the general
refusal vocabulary step 3 asks for. Steps 2 and 4 — at most one connection per peer
identity, and reaping a dead incumbent — and the duplicate-specific parts of 3, 6 and 7
are split out to GUM-339; the ticket itself says step 2 fixes the split outbox
independently of step 1, and the two properties review separately.

## The defect

A liaison server took the peer number a socket claimed at face value. Nothing was
signed, nothing was challenged. The hub's only guard was "is this a coil I hub" —
which any of the other 99 participants satisfies for any coil number — and the
mesh's was the dial-direction rule, which is topology, not identity.

## What changes

Both links now open with a three-frame exchange, and the **server speaks first**:

| # | frame | direction | carries |
|---|---|---|---|
| 1 | `Challenge(nonce)` | server → dialer | 32 random bytes, fresh per socket |
| 2 | `Handshake(num, protocolVersion, auth)` | dialer → server | the claimed identity and the proof |
| 3 | `Refused(refusal)` | server → dialer | why, immediately before the socket closes |

`auth` is a signature over `HandshakeProof.preimage`, checked against the roster key
the claimed number resolves to — `CoilPeers` on `/hub`, `HeadPeers` on `/head`. The
credential is the one the protocol is already built on, so there is no second
identity registry to drift.

```
handshakeProof = "gummiworm-handshake-v1"
  || <link>            // 0x01 coil→hub, 0x02 head↔head
  || <claimant>        // the coil or head peer number being claimed
  || <protocolVersion>
  || <headParamsHash>
  || <nonce>           // the server's per-socket challenge
```

Every term closes one replay, and every one has a test:

- **`nonce`** binds the proof to this socket. Without it a handshake captured once is
  good on any socket forever.
- **`link`** answers the ticket's open question 1 — a proof harvested on one lane is
  worthless on the other. Pinned by a test that sends head peer 0's own key, own
  number and this socket's nonce with a `CoilToHub` proof, and is refused.
- **`headParamsHash`** refuses both a peer from another head and one that disagrees
  about this one, at connect rather than on divergence. It subsumes the head id, and
  is already in the multisig regime datum, so it costs nothing to send.
- **`protocolVersion`** is #744's, carried into the proof so it cannot be edited in
  flight.

Checks run in a fixed order: version → roster or dial topology → proof. Version first
because a peer speaking another protocol may not mean the same thing by its own
number; the roster next because the claimed number is what resolves the key.

## A refused handshake now closes the socket, named

Previously the `else` branch only traced. The socket stayed open, the send stream
blocked forever on `coilD.get` — so no keep-alive pings flowed either — and the
connection died 20s later with no explanation, which reads as a network fault. Both
transports had the identical shape.

Now the server sends `Refused(refusal)` and a WebSocket close with status 1008 and
that reason. `HandshakeRefusal` keeps the six conditions distinct —
`ProtocolVersionMismatch`, `NotHubbed`, `NotInRoster`, `WrongDialDirection`,
`HeadParamsMismatch`, `BadSignature` — because each is a different operator action,
and collapsing them makes a key mix-up read like a misconfigured roster. The dialers
trace the refusal and **keep redialing**: a refusal describes the attempt, not the
peer.

## The dialers wait for the challenge

`WsDuplex.firstLine` collects the opening frame before the socket is claimed, bounded
at 10s. That budget stays under the 30s WebSocket-handshake budget on purpose: an
attempt that outlived the loop's claim would sit in `receive` holding an open socket
while the loop had already redialed. Both dialer-budget suites gained a case for it.

## The unproven form is gone

`HandshakeAuth` was #744's stub — an enum whose only case was `Unauthenticated`. It is
now a required product (head-params hash + signature), and
`HandshakeRefusal.Unauthenticated` went with it. No head is live, so nothing needs to
speak the unproven form, and a case that exists only to be rejected is a case to keep
in step.

One consequence, and the only handshake fault with no named refusal: a frame with no
`auth` now fails to decode, is traced as `ServerDecodeError`, and its socket lingers
to the idle timeout. Reachable only from a build predating this exchange, of which
there are none. Noted in the spec.

Note the asymmetry this leaves with `protocolVersion: Option[Int]`, which #744 made
optional for exactly the legibility reason this drops. Deliberately left as #744's
call rather than reversed here — worth settling one way or the other in review.

## Config changes

None. No config field is added, removed or renamed, and neither decoder changes, so
every existing `head-config.json` and `private.json` decodes unchanged. The
`create` signatures on the three transports grew parameters, but they are fed from
values `NodeConfig` already exposes (`ownWallet`, `headConfig`, `headParamsHash`).

## Wire compatibility

A deliberate break on top of #744's, and for the same reason: there are no live heads.
A peer that does not send a signed `Handshake` cannot open a link.

## Specs

`docs/spec/coil-network.md` §4.3 carries the exchange and the refusal vocabulary.
`docs/spec/head-params-hash.md` gains **check 5** — the handshake is now a place the
digest is compared, and that file's "what the head does not have is connection
authentication" paragraph was stale as of this PR.

## Checks

`just fmt`, `just lint`, `CI=true sbt "Test/compile; integration/Test/compile"` clean.
`testOnly *` 544 passed (517 on #744). `integration/testOnly *` 62 passed — the real
WS bring-up exercises the exchange end to end.

New suites: `HandshakeProofTest` (one test per preimage term; a malformed signature
refuses rather than throwing, as `platform.verifyEd25519Signature` does on a bad
length) and `HandshakeAcceptTest` (a real Ember server with frames driven by hand:
challenge-first, every refusal named on the wire *and* traced, nonce replay,
cross-lane replay).


euonymos:
Added: **the server announces its protocol version in the `Challenge` too**, so each side reaches its own verdict instead of waiting to be told.

This closes a gap rather than adding a feature. `design/versioning.md` (#738) has specified "both sides send it" since 2026-09-12; the code only ever had it in the dialer's `Handshake`, because when #744 added `protocolVersion` there was no `Challenge` — it arrived here, for the nonce, and nobody revisited the placement.

**Why it matters.** The dialer already learns both versions from `Refused(ProtocolVersionMismatch)`, so this is not a correctness fix. Two things are still bought:

1. The dialer refuses **before signing** a proof for a peer it cannot talk to.
2. The dialer can name both versions against a server that says nothing — and a server too old to send `Refused` is exactly that server, and exactly the population the version check exists to catch. Waiting to be told makes the diagnosis depend on the counterpart having the very mechanism under test.

**Shape.** `Challenge(nonce, protocolVersion: Option[Int])` on both envelopes, with `Challenge.own` alongside `Handshake.own`. `Option` for the same reason the handshake's is: a counterpart announcing none gets a legible refusal, not a decode failure. The check runs in both dialers before the handshake is built.

The `Challenge`'s version is **unsigned** — the server has no identity to prove until it has seen a claimant, and a forged version can only cause a refusal, which an attacker who can rewrite the frame could force anyway by dropping the socket.

New trace `DialerRefusedChallenge` on both lanes, distinct from `DialerRefused` (the counterpart refusing us) — the mirror direction, and the one that holds when the counterpart never explains itself.

**Tests** (+6): challenge round-trip asserts the announced version on both envelopes; a challenge with no `protocolVersion` decodes to `None` rather than failing; and on each lane a dialer facing a mismatched — and an absent — version refuses every attempt, opens no link, and keeps redialing.

`testOnly *` 550 green. `integration/testOnly *` 61/62 with one error in `SeedAbstractionTest`, which passes on its own and is rule-based-regime property work unrelated to the transports; this branch ran 62 green before the change, so I am treating it as a flake rather than a regression.

